PR #1, opened 2015-01-06
Update URLs to be mypage and use erightsid from request header

adgad:
Note: need to set header when developing locally, which could be annoying


matthew-andrews:
Also means we can make “going to my page” from header part of the core experience whoop whoop

server/app.js
```diff
@@ -30,11 +30,12 @@ app.use(require('./middleware/auth'));
  *  /users/<user-id>/<list> - Eg, following, history
  *  /groups/<group-id>/<list> - Eg, recommends
  */
-app.get('/users/:uid/following/new', require('./controllers/following'));
-app.get('/users/:uid/recommended', require('./controllers/recommended'));
```
matthew-andrews: Do we need to co-ordinate updating these across all the apps?  Perhaps 302 the old URL for at least `http://next.ft.com/users/:uid/following/new`?

adgad: Yup, I was planning to just coordinate since I think it's just the header
that has the link - but a redirect sounds like a good idea.

On 6 January 2015 at 17:41, Matt Andrews <notifications@github.com> wrote:

> In server/app.js
> <https://github.com/Financial-Times/next-ilyushin/pull/1#discussion-diff-22537317>
> :
>
> > @@ -30,11 +30,12 @@ app.use(require('./middleware/auth'));
> >   *  /users/<user-id>/<list> - Eg, following, history
> >   *  /groups/<group-id>/<list> - Eg, recommends
> >   */
> > -app.get('/users/:uid/following/new', require('./controllers/following'));
> > -app.get('/users/:uid/recommended', require('./controllers/recommended'));
>
> Do we need to co-ordinate updating these across all the apps? Perhaps 302
> the old URL for at least http://next.ft.com/users/:uid/following/new?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Financial-Times/next-ilyushin/pull/1/files#r22537317>.
>

-- 

------------------------------
This email was sent by a company owned by Pearson plc, registered office at 
80 Strand, London WC2R 0RL.  Registered in England and Wales with company 
number 53723.

adgad: 301? Moved permanently...

PR #4, opened 2015-01-13
Use elastic search and limit the 'seen' notifications we're showing

adgad:
/cc @wheresrhys @richard-still-ft


server/controllers/following.js
```diff
@@ -100,7 +105,11 @@ module.exports = function(req, res, next) {
 			};
 		});
 
-		var notificationIds = notifications.Items.map(function(item) {
+		var oneDayAgo = new Date().getTime() - (1000 * 60 * 60 * 24);
+		var notificationIds = notifications.Items.filter(function(item) {
+			return (item.Meta.S && !JSON.parse(item.Meta.S).seen) ||
+			(new Date(item.Created.S).getTime() >  oneDayAgo);
+		}).map(function(item) {
```
wheresrhys: Maybe raise an issue on user-prefs-api to accept date query string, but fine here for now

wheresrhys:
:+1:

PR #9, opened 2015-01-19
Fix numbers

adgad:
/cc @wheresrhys

- Number on nav shows count from more as well
- mark section as read clears more section
- if no notifications, still show the nav on the left


wheresrhys:
:+1:

PR #11, opened 2015-01-19
Fix numbers

adgad:
/cc @wheresrhys 

Date filter from user preferences api, rather than hacky filter inside controller.


wheresrhys:
:+1:

PR #15, opened 2015-01-20
Fix some encoding issues and the remove all button

adgad:
Hopefully makes the encoding consistent

/cc @wheresrhys


wheresrhys:
@adgad I just added a helper to to next-express to encode content. So worth switching to using that in future `{{encode thing}}`

PR #51, opened 2015-03-03
Stop trying to make fetch happen

adgad:
(if it doesn't happen within 3 seconds)

* Replace request with fetch everywhere
* Remove unused History pages
* Add timeout to all fetch calls


matthew-andrews:
yes yes yes!

PR #57, opened 2015-03-06
Sourcemaps

aintgoin2goa:


Makefile
```diff
@@ -53,10 +53,10 @@ debug:
 	export ELASTIC_SEARCH_URL=${ELASTIC_SEARCH_URL}; export HOSTEDGRAPHITE_APIKEY=123; export apikey=${API_KEY}; export PORT=${PORT}; node --debug-brk server/app.js
 
 build:
-	export ENVIRONMENT=development; gulp
+	export ENVIRONMENT=development; ./node_modules/.bin/gulp build-dev
 
 build-production:
-	gulp
+	./node_modules/.bin/gulp build-prod
```
matthew-andrews: Please remove ./node_modules/.bin - see notes on here http://financial-times.github.io/next/docs/developer-guide/ under common problems if that doesn't work for you

PR #80, opened 2015-03-19
Use _enqueueMicrotask branch of mustard to fix ios7

adgad:
/cc @matthew-andrews Makes the functionality work on ios7/android - but not the js-success tests, so will check that out next


matthew-andrews:
don't need to do this 2.2.2

adgad:
Oh cool, will update both

PR #93, opened 2015-03-26
Include OBT_DEBUG for some more timing information for the build.

samgiles:
The TaskLoader in `obt` outputs how long it took to load all the dependencies from disk, it would be good to see this information for a real build in CI so we can target performance improvements (and know if we're going in the right direction with dependency removal and simplification, this is gonna have to be little steps).

I've also removed a native dependency from the dependency tree so installation time should be significantly reduced now for v2.13.0.

Can we try it?


samgiles:
Decided against this.

PR #97, opened 2015-03-31
Don't use elements `dataset`

ironsidevsquincy:
Doesn't play well with IE<11

https://app.getsentry.com/nextftcom/ft-next-frontend/group/60535752/


matthew-andrews:
is this not polyfillable?

ironsidevsquincy:
Not according to http://polyfill.webservices.ft.com/v1/docs/features/#Element

matthew-andrews:
Right you are: @ironsidevsquincy  https://github.com/Financial-Times/polyfill-service/issues/188

matthew-andrews:
:+1:

PR #99, opened 2015-04-02
WIP: VERY quick first stab at an inbox-like my page

adgad:
/cc @wheresrhys 

- [x] delete all the backfilling article code
- [x] delete recommend/history pages
- [ ] use new templates from article cards
- [x] manage/delete follows
- [x] unsaving for later
- [ ] probably more


adgad:
I think this is probably good enough for an MVP.

No fancy styling yet but it's worth waiting till Style Guide a little further along with colors/typography to avoid duplicating effort.

client/components/topics/main.scss
```diff
@@ -0,0 +1,55 @@
+.mypage__topic__follow {
+	@include nextButtonsFollow();
+}
+
+.mypage__topic {
```
wheresrhys: While fancy UI is out it'd still be nice to alter these topics to have a remove icon, or text more strongly suggestive of removal

server/controllers/notifications-feed.js
```diff
@@ -0,0 +1,60 @@
+/*jshint node:true*/
```
wheresrhys: :+1: for splitting this out

server/lib/stream-utils.js
```diff
@@ -24,7 +24,7 @@ module.exports = {
 
 	formatSection: function (s) {
```
wheresrhys: what's this used for now?

wheresrhys:
good stuff

wheresrhys:
Should this be possible though 
image

server/controllers/notifications-feed.js
matthew-andrews: Pro Tip: if you use nbt verify rather than obt verify you won't need to set `/*jshint node:true*/ anymore

matthew-andrews: > if you use nbt verify
which [you do](https://github.com/Financial-Times/next-ilyushin/blob/master/Makefile#L17)

client/components/topics/main.scss
adgad: Yup, let's discuss this as part of the CTA work

adgad:
I think that might be to do with updated articles. We sort by/display last publish time (Rather than initial publish time) so that article was probably on your page previously, and then has since been updated hence appearing as 24 minutes ago.

wheresrhys:
I wonder if editorial have/cud have the ability to mark updates as
substantial, ten we could have lastSubstantiallyUdatedDate. In the mean
time we should ask mark/mus what proportion of updates are substantial. If
it's tiny we sort by firstpublished?

On Friday, May 8, 2015, Arjun Gadhia <notifications@github.com> wrote:

> I think that might be to do with updated articles. We sort by/display last
> publish time (Rather than initial publish time) so that article was
> probably on your page previously, and then has since been updated hence
> appearing as 24 minutes ago.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Financial-Times/next-ilyushin/pull/99#issuecomment-100160097>
> .
>

-- 

------------------------------
This email was sent by a company owned by Pearson plc, registered office at 
80 Strand, London WC2R 0RL.  Registered in England and Wales with company 
number 53723.

PR #106, opened 2015-04-23
Authenticate a users session and use this to get the userId

adgad:
/cc @wheresrhys @aintgoin2goa @commuterjoy


wheresrhys:
:+1:

PR #109, opened 2015-04-24
Logged in smoke tests

adgad:
/cc @matthew-andrews @spriggsha 

Close but no cigar - I can't get IE10 to pass :(


matthew-andrews:
we should definitely get it passing on ie10

.travis.yml
```diff
@@ -1,6 +1,6 @@
 sudo: false
 script:
-- make install && make test
+- make install && make test && make provision
```
matthew-andrews: need a `make tidy` method to throw away the VMs otherwise @richard-still-ft will get angry

matthew-andrews: https://github.com/Financial-Times/next-grumman/blob/eedb78053a7a8ed0e005c8a5d51e40583a536f36/Makefile#L84-85

PR #116, opened 2015-05-06
Endpoint to add/remove follows for core experience

adgad:
/cc @wheresrhys @matthew-andrews   This look okay for Core experience?


matthew-andrews:
nice!

adgad:
Going to make it a POST rather than a GET - so can be used as a form action.

adgad:
Handled by https://github.com/Financial-Times/next-user-preferences-api-v2/pull/71

PR #120, opened 2015-05-11
Fix a potential timeout

adgad:
/cc @wheresrhys 

I don't think this would be calling all the 503s, but...


wheresrhys:
My hunch is there's some article missing/missinga  property in the data returned. Anyone who happens to have that article on their mypage it breaks for them. This was the underlying reason for buggy releases I'd done before.

Are we using raven properly here? The sentry errors aren't very informative

PR #134, opened 2015-05-20
redirecting /mypage to appropriate start page

wheresrhys:
@adgad


wheresrhys:
annoyingly this endpoint still gets authorized, but don't want to mess around with middleware order right now

PR #141, opened 2015-05-21
Fix extra padding on streams

adgad:
/cc @wheresrhys @matthew-andrews


matthew-andrews:
There's a fix for this in OBT 3 Beta (ie. use lib sass rather than ruby sass)

PR #146, opened 2015-05-26
Clear notifications not on page

adgad:
/cc @wheresrhys Sometimes I get notifications which don't appear on my page (possibly because either the article gets deleted, or for some reason they don't get picked up by elastic search).

Changing code to clear these so that we don't get phantom red dots.


matthew-andrews:
ahh i think i've seen this^^

PR #147, opened 2015-05-26
Allow anon

aintgoin2goa:
Have done this the simplest way possible for now.  It doesn't seem to throw any errors as far as I can tell


.gitignore
```diff
@@ -12,3 +12,5 @@
 /npm-debug.log.*
 /tests/browser/reports
 /tests/browser/screenshots
+.editorconfig
+.idea/
```
wheresrhys: Would you mind adding these to your global .gitignore instead/as well? They're the kind of entries that may inadvertently get deleted when .gitignore is tidied up sometime in the future

wheresrhys:
:+1:

.gitignore
aintgoin2goa: Until this moment I had no idea there was a global .gitignore file, so thanks for that!  Have done it

PR #153, opened 2015-06-01
Remove email parameter from user prefs request

phamann:
We no longer need to send or save this data as the email job now requests the data on the fly for DAM.

/cc @adgad


adgad:
You'll also need to get rid of these: data-meta='{"email": "{{emailAddress}}"}'

phamann:
@adgad do you know what the data attributes are used for?

adgad:
Yup, so the data attributes are used by the JS. The hidden input was added for core experience.

PR #169, opened 2015-06-08
Optionally override the consumer of the feed

adgad:
/cc @wheresrhys


wheresrhys:
What d'you need this for - d'you mind leaving until after I've merged RSS?

PR #176, opened 2015-06-09
upgraded api client

wheresrhys:
@matthew-andrews @tom-parker 

I'd really like it to be possible to copy and paste the article-v1 model from dobi to ilyushin whenever there's a change, but for this to be possible I'd think it'd need to

a) support chaining in some sense
b) not overwrite any properties which already exist

thoughts?


tom-parker:
If we need to copy / paste it into ilyushin, wouldn't be better off in it's own module possibly? article-v1 basically deals with formatting / logic around the raw data to get it in the right state for the card?

Yes, would be good for it to not clobber properties that are already there. Not sure if they'll ever be a case when this is desirable though?

What chaining will it need?

wheresrhys:
Perhaps something a bit like the following would return a function that can then be passed into `Array.map`, that way the properties you want to manipulate could be mix and matched

```
var articleModel = require('article-model-v1');
myCardModelFactory = articleModel
   .defineHeading()
   .setVisualTone()
  .chain(function (article) {
     // some custom action I want to do
  })
  .setMedia(); 
// all the above returns a function

var cardModels = myArticles.map(myCardModelFactory)
```

Could be a way to share the useful stuff without making any one-size-fits all assumptions

tom-parker:
:+1:  that seems like a really nice solution. Also gives us the right amount of flexibilty whilst still keeping the set heading / set visual tone / set media which all the cards need.

PR #203, opened 2015-07-14
relying on accurate all tags list coming back from elastic search

wheresrhys:
Falls back to just missing some articles when elastic search is off @matthew-andrews @adgad


matthew-andrews:
:)

PR #205, opened 2015-07-15
No genre distinction on feed

adgad:
/cc @wheresrhys 

For the sake of consistency, this will affect the tray and the reading list


wheresrhys:
:+1:

PR #216, opened 2015-07-21
Redirect unauthorized users to homepage

adgad:
/cc @wheresrhys 
Currently when the user logs out from the myFT page, they get taken to an error page (not sure if there's a nicer way aroudn this than redirect to HP?)


wheresrhys:
presumably we can design a page and redirect to there eventually, but homepage better than error :+1:

PR #219, opened 2015-07-23
WIP: Push Notifications

adgad:
/cc @matthew-andrews @phamann @wheresrhys @commuterjoy 

**myFT page**
- [x] Add service worker
- [x] Add push notification signup buttons
- [x] Register with browser push & GCM
- [x] Display notification on recieve push
(Can verify up to this point by sending pushes via curling https://android.googleapis.com/gcm/send)
- [x] Keep GCM and UI state in sync with user-prefs (need to store GCM)
- [x] ~~Store which topics to recieve push for (either a property of push preference or myFT topics)~~ - For now just try it internally with all topics since that's easier

**Content Notifier**
- [x] New article -> Get users with push  -> Check if article matches some criteria -> Send GCM push
- [x] Format/content of actual notification (title, img?, URL)

**General**
- [ ] UX around multiple devices: i.e. if I sign up for push on browser, should I get on mobile also?
- [ ] SW currently in ilyushin (so only gets registered once user visits myFT) - fine whilst signup to push on myFT but maybe needs to be moved more global if myFT becomes less of a destination.
- [ ] Shared Next Google account for GCM (was using my email add to test)


client/components/topics/index.js
```diff
@@ -20,22 +22,28 @@ module.exports = {
...
-			useConceptId: flags.get('userPrefsUseConceptId'),
-			searchEl: document.querySelector('.mypage__topic-search input[type="search"]'),
-			destinationEl: topicsListEl,
-			resultsMessageEl: document.querySelector('.mypage__topic-search__results-title'),
-			resultsMessage: 'Results:',
-			resultTpl: document.querySelector('.mypage__topic-suggestions__template').textContent,
-			noResultsTpl: '<li class="mypage__topic-suggestions__no-topics">No topics matching <b>{searchTerm}</b></li>',
-			maxResults: 3
-		});
+		if(flags.get('topicSuggestions')) {
```
adgad: Oops, snuck this in here because the topicSuggestion flag expiry was causing some JS errors!

server/app.js
```diff
@@ -43,6 +45,15 @@ app.use('/mypage', legacyRouter);
 
 var router = express.Router();
 
+router.get('/service-worker-push.js', function(req, res, next) {
+	res.set('Cache-Control', 'private, max-age=0, no-cache');
+	res.sendFile(require('path').resolve(__dirname + '/../client/service-worker-push.js'));
+});
+
+router.get('/manifest.json', function(req, res, next) {
```
adgad: Will remove and add a new manifest in next-assets

views/partials/notification-preferences.html
```diff
@@ -44,8 +44,17 @@ <h3 class="mypage__subtitle">Global alert settings</h3>
...
+<div class="myft-ui myft-ui--prefer"
+	data-preference-name="push-notifications">
+	<button
+		type="submit"
+		class="myft-ui__button"
+		data-trackable="set-{{preferenceName}}" disabled
+	><span>Subscribe</span></button>
+	<label>Get push notifications</label>
+</div>
+{{/if}}
```
adgad: Note: had to not use the template because the handling of button state/sending stuff to server is quite coupled to whether push is actually supported/getting GCM ids - so bypassing next-myft-ui Javascript.

client/service-worker-push.js
```diff
@@ -0,0 +1,85 @@
...
+
+	var title = 'New article in your myFT page';
+	var tag = 'next-myft-article';
+	var icon = 'https://next-geebee.ft.com/image/v1/images/raw/http%3A%2F%2Fnext.ft.com%2Fassets%2Ficons%2Fmyft-bolder.svg?source=next&tint=%23c00d00,%23c00d00&format=png&width=48';
+
+
+	event.waitUntil(
+
+		fetch('/myft/my-news.json?since=-1h', {
+			headers: {
```
adgad: This won't work for logged out users - but might not be an issue? (falls back to generic text on error)

client/components/topics/push-notifications.js
```diff
@@ -0,0 +1,221 @@
...
+		} else {
+			subscribe();
+		}
+	});
+
+	if ('serviceWorker' in navigator) {
+		navigator.serviceWorker.register('/myft/service-worker-push.js')
+		.then(initialiseState);
+	} else {
+		console.warn('Service workers aren\'t supported in this browser.');
```
phamann: Maybe disable or remove the push notification button/ui at this point?

client/components/topics/push-notifications.js
```diff
@@ -0,0 +1,221 @@
...
+	}
+}
+
+// Once the service worker is registered set the initial state
+function initialiseState() {
+	// Are Notifications supported in the service worker?
+	if (!('showNotification' in window.ServiceWorkerRegistration.prototype)) {
+		console.warn('Notifications aren\'t supported.');
+		return;
```
phamann: As above.

client/components/topics/push-notifications.js
```diff
@@ -0,0 +1,221 @@
...
+				// Keep your server in sync with the latest subscriptionId
+				sendSubscriptionToServer(subscription);
+
+				// Set your UI to show they have subscribed for
+				// push messages
+				nButtons.toggleState(pushButton);
+				isPushEnabled = true;
+			})
+			.catch(function(err) {
+				console.warn('Error during getSubscription()', err);
```
phamann: Do we have our CS Sentry configured to track `warn`

client/components/topics/push-notifications.js
```diff
@@ -0,0 +1,221 @@
...
+		serviceWorkerRegistration.pushManager.subscribe({userVisibleOnly: true})
+			.then(function(subscription) {
+				// The subscription was successful
+				nButtons.toggleState(pushButton);
+				isPushEnabled = true;
+				pushButton.disabled = false;
+
+				// TODO: Send the subscription subscription.endpoint
+				// to your server and save it to send a push message
+				// at a later date
```
phamann: Is this comment needed?

client/components/topics/push-notifications.js
```diff
@@ -0,0 +1,221 @@
...
+				if (Notification.permission === 'denied') {
+					// The user denied the notification permission which
+					// means we failed to subscribe and the user will need
+					// to manually change the notification permission to
+					// subscribe to push messages
+					pushButton.disabled = true;
+				} else {
+					// A problem occurred with the subscription, this can
+					// often be down to an issue or lack of the gcm_sender_id
+					// and / or gcm_user_visible_only
```
phamann: Would be good to keep track of how many times subscriptions fail at this point. Maybe beacon to Sentry or Keen?

client/components/topics/push-notifications.js
```diff
@@ -0,0 +1,221 @@
...
+					// to allow the user to subscribe to push
+					isPushEnabled = false;
+					pushButton.disabled = false;
+					nButtons.toggleState(pushButton);
+					return;
+				}
+
+				// TODO: Make a request to your server to remove
+				// the users data from your data store so you
+				// don't attempt to send them push messages anymore
```
phamann: Can remove as you are doing below?

server/app.js
```diff
@@ -43,6 +45,11 @@ app.use('/mypage', legacyRouter);
 
 var router = express.Router();
 
+router.get('/service-worker-push.js', function(req, res, next) {
+	res.set('Cache-Control', 'private, max-age=0, no-cache');
+	res.sendFile(require('path').resolve(__dirname + '/../client/service-worker-push.js'));
```
wheresrhys: So do we now need to update nbt to support multiple js bundles. Or does it already @matthew-andrews ?

wheresrhys: also hashed-assets

phamann: The `private, no-cache` will solve the hashing issue for the time being, but yes we need to support bundling soon as we'll want to start `require()` deps inside these files.

phamann:
We also need to consider our service worker upgrade path, as i'm sure we'll want to tweak this file going forward.

adgad:
From what I understand Browsers handle this themselves - so it will download the (uncached) file in the background and if there's a byte difference will install in the background. Not sure how well it works in practice.

PR #220, opened 2015-07-24
parallelization of auth and content fetching

wheresrhys:
@matthew-andrews 

Adding you as reviewer in absence of Patjun

Basic idea - If we drive users to urls that contain their uuid we can then authorize in parallel with content fetching and drive response times down by this much http://graphite.ft.com/render/?width=586&height=308&_salt=1437732140.078&target=averageSeries(next.heroku.ilyushin.*.fetch.session.response.status_2xx.response_time.max)&target=averageSeries(next.heroku.ilyushin.*.fetch.session.response.status_2xx.response_time.mean)&from=-48hours ... queue fewer timeouts and 503s

This to be followed by client side work to rewrite urls in links once user data is fetched


server/middleware/auth.js
```diff
@@ -48,20 +50,31 @@ module.exports = function (req, res, next) {
 		}
 		var headers = req.headers;
 		delete headers.host;
-		fetch('https://session-next.ft.com', {
+
+		res.locals.authPromise = fetch('https://session-next.ft.com', {
```
matthew-andrews: timeout?

PR #234, opened 2015-08-03
Use block card type (vertically stacked) for tray content

keirog:


keirog:
Before | After
--- | ---
before | after

PR #239, opened 2015-08-06
Fix compatibility issues with new o-grid

adgad:
/cc @ironsidevsquincy @i-like-robots (Don't have IE8 here yet but does this look about right?)


wheresrhys:
:+1:

PR #240, opened 2015-08-06
middleware not redirecting root

wheresrhys:
So that it's easier to write the redirects on the client side and also making way for the dashboard.

Maintenance mode is a bit smarter too


server/app.js
```diff
@@ -24,8 +24,10 @@ app.get('/__gtg', function(req, res, next) {
...
 
-legacyRouter.use(require('./middleware/root-redirect'));
-legacyRouter.get('/', function (){});
+// dummy controller to force the redirect middleware to run
+legacyRouter.get('/', function (req, res) {
+	res.redirect(302, '/myft');
+});
 legacyRouter.get('/following', function (req, res) {
 	res.redirect(302, '/myft/my-news');
 });
```
adgad: Whilst we're at it could we get rid of some of these legacy routes?

wheresrhys: I'll add a metrics counter to see if they're getting hit much. If not then remove later

PR #246, opened 2015-08-10
Don't send http errors to sentry

matthew-andrews:
@adgad @wheresrhys

PR #253, opened 2015-08-11
Redirect to the old portfolio page on error

adgad:


server/controllers/portfolio.js
```diff
@@ -45,6 +46,9 @@ module.exports = function(req, res, next) {
 			res.render(template, viewModel);
 
 		})
-		.catch(next);
+		.catch(function(err) {
+			errorHandler.captureMessage('Redirecting to old portfolio', err);
```
matthew-andrews: eek… not an error!  log to splunk or use metrics pls

matthew-andrews: e.g. https://github.com/Financial-Times/next-ilyushin/pull/246/files

commuterjoy:
I think this is confusing. Would be better just to serve an error, explain why, and then ask them to refresh.

server/controllers/portfolio.js
adgad: It would have been logging an error previously anyway no? I can check for timeouts and not error then, but the rest should probably still go to sentry

adgad:
I'll hold off the redirect and add the better timeout logging so we can see if it is actually a problem. Also tweaking the fetch timeout so will hopefully improve.

PR #267, opened 2015-08-19
Portfolio dns

adgad:
/cc @phamann @wheresrhys @richard-still-ft @commuterjoy 

If flag turned on, resolve portfolio DNS ourselves and fetch that.

(Also adding copy that James wanted on top of Portfolio, moving edit link and adding back percentages)


wheresrhys:
:+1:

PR #281, opened 2015-08-27
More color tweaks and cleanup

adgad:
/cc @pauloneillft @andygnewman 
screen shot 2015-08-27 at 09 30 51


pauloneillft:
Looks good. I assume the hover is the same as selected.

PR #288, opened 2015-09-04
Product tour

Lrnk:
Product tour in progress. A bit rough, including temporary images. Accessible from myft/product-tour only (no links to it anywhere yet).

By default, slices that we don't want to see yet are hidden (hui, prefs, device sync), but the entire tour can be accessed using this query string: myft/product-tour?slices=standfirst,follow,alerts,push,save,hui,portfolio,prefs,sync,testimonials


client/components/product-tour/index.js
```diff
@@ -0,0 +1,140 @@
+"use strict";
+
+var Tweenable = require('shifty');
+
+module.exports = {
+	init: function (flags) {
+
+		const hasAnimated = Symbol();
```
adgad: Does this work in IE? I think polyfill-service doesn't yet polyfill Symbol...

client/components/product-tour/main.scss
```diff
@@ -0,0 +1,340 @@
+.myft-tour__slice {
+
+	//bit of a hack to make sure parallax scrolling bits don't bleed out of slices
+	transform: translate3d(0, 0, 0);
+
+	background-color: oColorsGetPaletteColor(warm-5);
```
adgad: You can use [use-cases](https://github.com/Financial-Times/next-sass-setup/blob/master/_functions.scss#L9) here (assuming it's supposed to be the same colour as the page background), so any page background changes carry through:

`getColorFor('page', 'background')`

https://github.com/Financial-Times/next-sass-setup/blob/master/_colors-use-cases.scss

client/components/product-tour/main.scss
```diff
@@ -0,0 +1,340 @@
...
+	p {
+		@include nTypeAlpha(3);
+	}
+
+	a {
+		@include nLinksBody();
+	}
+
+	strong {
+		color: oColorsGetPaletteColor(claret);
```
adgad: Possibly same here, to be topic use case

server/middleware/auth.js
```diff
@@ -66,9 +66,13 @@ var authWithSessionToken = function (urlUserId, req, res, next) {
 module.exports = function (req, res, next) {
 	var urlUserId = (urlUserIdRX.exec(req.path) || [])[1];
 
+	// Allow anonymous access to product tour page
+	if (res.locals.anon.userIsAnonymous && /^\/myft\/product-tour/.test(req.originalUrl)) {
+		next();
+
```
adgad: Should this be just anonymous, or everyone?

views/product-tour.html
```diff
@@ -0,0 +1,20 @@
...
+	{{{usePartial 'product-tour-slices/alerts'}}}
+	{{{usePartial 'product-tour-slices/push'}}}
+	{{{usePartial 'product-tour-slices/save'}}}
+	{{!-- {{{usePartial 'product-tour-slices/hui'}}} --}}
+	{{{usePartial 'product-tour-slices/portfolio'}}}
+	{{!-- {{{usePartial 'product-tour-slices/prefs'}}} --}}
+	{{!-- {{{usePartial 'product-tour-slices/sync'}}} --}}
+	{{{usePartial 'product-tour-slices/testimonials'}}}
+
+{{/if}}
```
adgad: Maybe neater if you have a default slices array in the viewmodel (without hui etc) that is overridden if you have the query params.

PR #294, opened 2015-09-10
Dashboard view MVP

adgad:
/cc @Lrnk @phamann 
* Currently replacing feed view (behind a flag) until we figure out what it is
* Logic for which cards to show is also non-existent at the moment (currently just shows all topics in a massive grid)
* BUT there are tests (shamelessly ripped off from stream page)


server/controllers/dashboard.js
```diff
@@ -0,0 +1,60 @@
...
+				viewModel.cards = [];
+			} else {
+				viewModel.cards = getTopicsFromArticles(content);
+			}
+			res.render('dashboard', viewModel);
+
+		})
+		.catch(function(err) {
+			if (err.name === fetchres.BadServerResponseError.name
+				|| err.message.indexOf('timeout') > -1) {
```
Lrnk: Can't resist the opportunity to suggest using ES6's new string functions: http://es6-features.org/#StringSearching
(compatible with node 4.0)

PR #298, opened 2015-09-10
Fix redirect for core experience

adgad:
/cc @matthew-andrews @phamann

Save button was working by fluke since Article IDs have the same pattern as User GUIDS

Ironically I was just starting tidying and adding tests for all the auth/redirection anyway (because making a change there was scary).


matthew-andrews:
go go go :)

matthew-andrews:
Do we know how long this has been broken for?

adgad:
Will look into it - but I suspect a couple of weeks, during the switch to the new API proxy

PR #301, opened 2015-09-14
Use api client

phamann:
This is a big PR which is mainly @Lrnk's great handy work.

- Wrap all calls to data APIs in the `myftApiV2` switch to toggle between source
- When switch is true use the new isomorphic `myFtApi` client to request data
- When switch is false map the old data structure to the new one
- Update API proxy
- Tests

/cc @adgad @Lrnk @wheresrhys @keirog


server/controllers/my-news.js
```diff
@@ -31,17 +32,32 @@ module.exports = function(req, res, next) {
 	} else {
 		// otherwise it's a html page - check whether the user is following something before
 		// deciding whether to redirect
-		redirectPromise = fetch('http://ft-next-api-user-prefs-v002.herokuapp.com/activities/User:guid-' + userId + '/followed/Topic:', {
-			timeout: 3000,
+
+		var topicsPromise;
+
+		// if (res.locals.flags.myFtIsomorphicApiClient) {
+		if (true) {
```
adgad: :scream_cat:

server/controllers/my-topics.js
```diff
@@ -18,13 +18,13 @@ function resultMessage(success, verb, conceptName) {
 }
 
 function fetchUserPrefs(type, userId, verb, subjectType) {
-	loadTestMyftApi.get(userId, verb);
```
adgad: If we turn the load test off, will we lose some data consistency between the two? (If the flag is off)

server/lib/api-proxy.js
```diff
@@ -1,25 +1,51 @@
...
-								opts.relationship === 'used' ? 'Feature:' :
-								opts.relationship === 'forlater' ? 'Article:' :
-								opts.relationship === 'preferred' ? 'Preference:' : '';
+module.exports = function (opts, flags) {
 
-	if (opts.subject) {
-		subject += opts.subject;
-	}
+	// if (flags.myftApiV2) {
+	if (true) {
```
adgad: :boom:

server/controllers/my-topics.js
adgad: Oh never mind, just realised this is the get!

phamann: Correct, only GET. But does raise another question that we'll most likely need to perform the mutating request in the opposite direction (i.e. still POST to user-prefs) when switch is on.

phamann:
Bar the comments above re. syncing this is now ready to ship.

server/controllers/my-topics.js
phamann: I've now resolved this issue inside the proxy: https://github.com/Financial-Times/next-ilyushin/blob/use-api-client/server/lib/api-proxy.js#L53

PR #303, opened 2015-09-16
Temporarily map api PUT's to POST's

phamann:
The old version of the API incorrectly used PUTs to create relationships instead of a POST (breaking fundamental principles of REST :cry:). 

The new API database query for PUTs its written to only match existing relationships and update their parameters and thus fails when used to create new relationships. 

This PR temporarily maps the PUT requests to POSTs.

/cc @adgad


adgad:
:+1:

PR #305, opened 2015-09-16
Tour updates

keirog:
@Lrnk Are you ok if I merge this branch into master? I'd like to start having smaller branch/PR changes, so we don't end up too far away from the live version.


Lrnk:
+1

PR #320, opened 2015-09-21
Add proxy for new API user call

adgad:

(only works for new API since old API doesn't have an equivalent)

Beacon can then use this endpoint and benefit from some caching.


Lrnk:
:+1:

PR #321, opened 2015-09-22
Fix RSS auth and urls

adgad:
//cc @Financial-Times/next-myft


test/server/auth.spec.js
```diff
@@ -40,12 +40,12 @@ describe('Authentication', function() {
 			{
 				name: 'rss-enabled',
 				matcher: '^https://ft-next-myft-api.herokuapp.com/v1/user/aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee/preferred/publish-rss-feeds',
-				response: { items: ['yes'] }
+				response: { uuid: "publish-rss-feeds" }
```
wheresrhys: I was thinking we should rename this to 'publish-news-feed-rss', which will allow for finer granularity when sharing saved article lists separately. 

As it's broken at the moment anyway it could be a good opportunity to change it without too much worrying

wheresrhys: Also, what do we think of changing the url to '/news-feed' as a clearer identifier when we roll out the dashboard?

adgad: Hmm. Might need to rethink how we do this auth actually, with sharing. It doesn't make sense for a user to have to enable publishing in order to have to share - so it might be that we have to create a new DB relationship e.g shared?

wheresrhys: hmmm, yeah maybe. We're gonna need a bigger relationships diagram

PR #328, opened 2015-09-24
upgraded to nbt 4 and did eslintiing

Lrnk:


client/components/topics/index.js
```diff
@@ -1,3 +1,4 @@
+/* global document */
```
matthew-andrews: we should update our eslintrc file to include this, the old jshintrc had this

matthew-andrews: also there should be an option in eslint to include browser variables rather than have to make this a `global`

client/main.js
```diff
@@ -1,4 +1,5 @@
-/*jshint node:true*/
+/* global window */
+/* eslint-env node */
```
matthew-andrews: again this should be handled with `eslintrc` imho

client/worker.js
```diff
@@ -1,4 +1,4 @@
-/* global self:false, clients:false */
+/* global self:false, clients:false, Notification */
```
matthew-andrews: should be covered by browser variables, in the central NBT eslintrc

client/components/preferences/push-notifications.js
```diff
@@ -1,3 +1,4 @@
+/* global window, document, navigator, Notification */
```
matthew-andrews: again, shoudn't need to add these — we need to add to our central eslintrc file…… `browser: true`
http://eslint.org/docs/user-guide/configuring

matthew-andrews:
Sorry to put loads of comments on this… we should aim to fix our central `.eslintrc` file to minimise the impact across the site………

Lrnk:
Yes I think you're right. I see you've already made a new NBT release so I'll move to that one and revert this stuff.

Lrnk:
Closing because the linting has changed and it also ended up containing a load of unnecessary commenting out of smoke tests etc

PR #333, opened 2015-09-28
Rename ilyushin to myft-page

phamann:
il2_sturmovik

@Financial-Times/next-myft


wheresrhys:
we should also aim to popularise pronouncing it 'miffed'

PR #343, opened 2015-10-02
redirect to old product tour whilst new one is broken

Lrnk:
:disappointed:


phamann:
:+1:

server/controllers/portfolio.js
```diff
@@ -15,6 +15,11 @@ function getPortfolioUrl(id, passportId) {
 }
 
 module.exports = function(req,res,next) {
+
+	// FIXME: temporarily redirect to old portfolio page while we fix portfolio
+	res.redirect(302, 'http://portfolio.ft.com/dashboard/');
```
matthew-andrews: better to use http://www.ft.com/marketsdata/portfolios because that'll send people to the mobile portfolio on m.ft.com and the normal one on other devices

Lrnk: Ah sorry only just spotted this! Will change.

PR #347, opened 2015-10-05
Use a URL that works for m.ft.com refugees as well as www. converts

matthew-andrews:


matthew-andrews:
Never mind, this is all completely broken now anyway.

adgad:
What's completely broken?

PR #351, opened 2015-10-06
Use new API for push notifications registration

adgad:

Use the `enabled`->`endpoint` relationship to store push notification preferences.

Depends on - https://github.com/Financial-Times/next-myft-client/pull/51


phamann:
:+1:

PR #354, opened 2015-10-08
no --skip-logging

matthew-andrews:


matthew-andrews:
Did you get SalesForce access yet @wheresrhys

matthew-andrews:
clearly this is completely broken.

PR #357, opened 2015-10-09
Refactor my-news logic to support grid and list views

phamann:
As discussed new logic
- Both grid and list view and now served via the same route `/myft/my-news`
- If `layout=` query param exists then it takes precedence
- If `myftDashboard` switch is enabled then it `grid` is always shown
- Otherwise users preference is used
- Defaulting to `list` if none of the above applies

I've also refactored the view data model slightly so each layout type can decorate the model with its needs.

Next step is to add the toggling UI.


adgad:
LGTM

server/controllers/my-news.js
```diff
@@ -33,24 +33,21 @@ module.exports = function(req, res, next) {
...
-		redirectPromise = topicsPromise
-			.then(function(hasTopics) {
-				shouldShowTour = res.locals.flags.myftProductTour && !hasTopics;
-				return !hasTopics;
-			})
-			.catch(function() {
-				return true;
-			});
+		redirectPromise = myFtApi.getAllRelationship('user', userId, 'followed')
+							.then(() => true, err => {
```
wheresrhys: why extra whitespace?

server/lib/view-decorators/feed.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
```
wheresrhys: rename this file to fragment?

wheresrhys:
:+1:

PR #360, opened 2015-10-09
Fix tray

phamann:
Fixes the tray layout since #357 


adgad:
Maybe we should just make the tray use ?layout=feed rather than ?fragment=true ?

phamann:
> Maybe we should just make the tray use ?layout=feed rather than ?fragment=true ?

@adgad Yes I thought this when writing this too. And will do in a separate PR, just that will require a PR into next-myft-ui too.

PR #363, opened 2015-10-12
Accessing myFT when anonymous goes to product tour, not home page

Lrnk:


wheresrhys:
:+1: 

First time I've seen
image

Can't help thinking it should be more noticeable

Lrnk:
Yeah perhaps, I'll put it on that doc I made of things to come back to on the product tour

PR #365, opened 2015-10-12
added smoke tests for unengaged & engaged users

wheresrhys:


adgad:
:+1: Is it worth adding some POST/DELETEs to test URLs too (for the engaged user)

PR #368, opened 2015-10-13
Grid view

phamann:
layout-switching


views/partials/toolbar.html
```diff
@@ -0,0 +1,25 @@
+{{#if @root.flags.myftDashboard}}
+<div class="myft-toolbar">
+	<div class="o-grid-container">
+		<div class="o-grid-row">
+			<div data-o-grid-colspan="12">
+
+				<div class="o-buttons-group myft-layout-switcher">
```
adgad: Should this be using next-buttons? Or has that been deprecated now (/cc @keirog )

wheresrhys:
I wonder should we bite the bullet and redirect `/myft/my-news?layout=whatever` to `/myft/my-news` once the preferences request returns.
cons - makes the change of layout a good deal slower
pros - avoids hitting myft-api on each subsequent page refresh, with the impact that woudl have on cache purging

Or is the response cached in the CDN, avoiding a second trip to the server?

Alternative might be to rewrite `window.location` client side if it contains a `layout` qs

adgad:
:+1: Plus tests for the new layout would be good :)

client/components/toolbar/main.scss
```diff
@@ -0,0 +1,48 @@
...
+		left: $spacing-unit/2;
+		top: 25%;
+	}
+
+	&:last-child {
+		margin-left: 5px;
+	}
+}
+
+.myft-layout-switcher__btn--list {
```
keirog: Maybe just have these styles in __btn, rather than as separate modifiers

views/partials/toolbar.html
keirog: I think this use case is fine to stick with o-buttons.

next-buttons only really exists for:
1. toggleState() for ajaxy buttons (which maaaybe could just live in myft-ui)
2. overriding o-buttons' font sizes to match our font matrix, but matrix fonts/sizes are being reviewed at the moment anyway (which should result in size parity between origami/next)
3. deprecated styles that we still need to phase out

client/components/toolbar/main.scss
phamann: @keirog I'm confused, the shared styles are in `__btn` but we need modifiers for each icon type?

PR #370, opened 2015-10-14
New caching logic

phamann:
- Tidy up the `shouldCache(req)` logic to be more terse/compose-able
- Increase `stale-if-error` time to 1 day
- Change `Surrogate-Key` value to be `uuid:page` to match new purging logic inside api (see: [#78](https://github.com/Financial-Times/next-myft-api/pull/78) for more detail)
- Add tests for new cache logic


server/middleware/cache-control.js
```diff
@@ -2,30 +2,34 @@
 
 const myFtApi = require('next-myft-client');
 
+function isRss(req) {
+	return (req.accepts('application/rss') || /\.rss$/.test(req.path));
```
Lrnk: this will break if there's a query string

phamann: `req.path` in this instance is just the path sans the query, as it's already been parsed.

Lrnk:
:+1: cool

PR #372, opened 2015-10-15
Portfolio currency selector

Lrnk:
Big commit for portfolio currency selection. Portfolio still redirects to portfolio.ft.com unless you've got the myftPortfolio flag set (we still need this workaround while they set a CDN up on the webservice)

It features the first instance within myft-page of sharing a handlebars helper and template between client and server, hence the new *isomorphic* directory, sitting alongside the *client* and *server* directories

It also adds a new endpoint portfolio-data (better name suggestions heartily welcomed) for getting the portfolio data in a lump of JSON and compiling the template on the client. 

It looks like this:

screen shot 2015-10-15 at 15 06 41


Lrnk:
going to have a go at following @wheresrhys's suggestion of using next-handlebars on the client side as opposed to vanilla handlebars

bower.json
```diff
@@ -15,7 +15,8 @@
     "o-expander": "^2.0.0",
     "o-forms": "^2.0.0",
     "o-table": "~1.7.2",
-    "shifty": "~1.5.0"
+    "shifty": "~1.5.0",
+    "handlebars": "~4.0.3"
```
keirog: :police_car: s/~/^

Lrnk:
No longer going to bring in n-handlebars as the version number isn't managed by this app, so it doesn't give us the peace of mind of knowing that everything that works on server-side handlebars will work on client-side handlebars after all. We can always bring it in when we need it.

PR #373, opened 2015-10-15
[WIP] Lists

adgad:

Really ugly and basic and not ready for public, but working enough to be able to test flows.
*/user/lists returns an unstyled list of that users lists
* /list/<listid> just uses saved articles templates for now
* /list</listid> is completely unauthenticated so anyone can see it

Also refactored the authentication a little, to avoid a nasty if statement in auth.js


Lrnk:
:+1:

PR #382, opened 2015-10-20
Use createReleationship instead of update for layout-switching

phamann:


adgad:
Isn't it `addRelationship`?

adgad:
Also `addRelationship`'s body is the node e.g. 
```
{ uuid: 'myft-layout' }
```

Not sure if we ever got round to adding support for this in the API, but you'd need to do something like:
```
{ 
uuid: 'myft-layout',
_rel: {
layout: 'grouped'
 }
```

phamann:
@adgad Good catch :wink: seems that the terminology in the `crud.js` does not match the client.

phamann:
@adgad Yep we do support `_rel` https://github.com/Financial-Times/next-myft-api/blob/master/server/v1/models/crud.js#L277-L291

PR #383, opened 2015-10-21
Create and add content to lists

Lrnk:

Core journeys for the following:
 - create list (from lists page only)
 - add article to list after saving (currently core experience only)

Ugly and not ready for public consumption, but functional. Lists page and 'add to list' form both hidden behind myftLists flag.


server/controllers/api/gateway.js
```diff
@@ -42,6 +42,8 @@ function humanReadableAction(method, relationship, subject, requestBody) {
 					return 'You have unsubscribed from receiving push notifications on this device' + settingsLink;
 			}
 			break;
+		case 'post:contained' :
+			return `You have added the article to the list. <a href="/myft/list/${subject}">View list</a>`;
```
wheresrhys: As a genral point all these confirmation messages could do with some copywriter attention maybe. I made most of them up. @james.webb

server/controllers/api/proxy.js
```diff
@@ -8,12 +8,24 @@ module.exports = function (req, res, next) {
 		res.sendStatus(400);
 		return;
 	}
+
+	let actor;
+	let actorId;
+	if(req.path.includes('core/list')) {
```
wheresrhys: Is there someway to keep this DRY? e.g. lib/getApiActor.js or something

server/controllers/lists.get.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
```
wheresrhys: :+1: for the file naming

views/partials/article-saved.html
```diff
@@ -0,0 +1,30 @@
+{{! todo this will move to next-myft-ui so it can be used in the enhanced experience }}
+{{! todo tracking}}
```
wheresrhys: ... no time like the present :simple_smile:

server/controllers/api/proxy.js
Lrnk: yep that sounds sensible I'll do that

server/controllers/lists.get.js
Lrnk: glad you approve

views/partials/article-saved.html
Lrnk: @wheresrhys The tracking you mean? yes you're right I should do that now really.

The next-myft-ui thing is more of a reminder to remove it when I've got the enhanced experience in. I'm unwilling to put it in next-myft-ui now because that means doing a release of that component for this, which is currently a bit naff.

Unless you thing I should do and point it to the branch?

server/controllers/api/gateway.js
Lrnk: I've added a card to the backlog

PR #384, opened 2015-10-21
myft_api_url env var at the start of each server test. also some linting

Lrnk:
Put MYFT_API_URL environment variable at the top of each server test file.
(also did some es6 linting while I was at it)

Now that myft api url comes from config vars, we need it in the tests. We need to decide which approach we prefer:
 - put the environment variable at the top of each test file (this PR)
 - put the environment variable in CI config vars (fixes tests on circle but not on our machines unless we set the env variable on each one)
 - some other solution I've not thought of

Any thoughts? /cc @Financial-Times/next-myft


wheresrhys:
Make is your friend here. put `export MYFT_API_URL=whatever;`  at the start of this line https://github.com/Financial-Times/next-myft-page/blob/master/Makefile#L17

Lrnk:
Ah that's loads better!

PR #390, opened 2015-10-28
Rm news feed log

Lrnk:
Stops a harmless but irritating 400 error that was happening on every single load of the news feed.


keirog:
:+1:

PR #392, opened 2015-10-29
made product tour and my-topics work with new concepts API

Lrnk:
a bit sad about losing the stream pun


server/jobs/hot-topics.js
```diff
@@ -1,40 +1,37 @@
...
+const topBrands = require('../models/top-brands');
+const topAuthors = require('../models/top-authors');
+
+let hotTopics = {};
+
+//adapt for new concepts API results schema
+function adaptTags (rawTags) {
+	return rawTags.map(rawTag => ({
+		id: rawTag.idV1 || rawTag.id,
+		name: rawTag.prefLabel || rawTag.name,
```
adgad: Is there any chance this would get a weird mix of v1 ID and v2 name? (which in turn means MERGEs on the DB would fail?)

adgad:
LGTM

server/jobs/hot-topics.js
Lrnk: Ah yeah that's an interesting consideration. There's no chance of this occurring with the APIs as they are currently, so as long as its versioned properly, we should be ok

Lrnk:
thank you!

PR #396, opened 2015-11-02
Save to list enhanced

laurieboyes:
The myft-page side stuff for the 'add article to list' enhanced pop-up experience. All user-facing stuff is behind myftLists flag.


server/controllers/article-saved.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
+const getLists = require('../lib/lists');
+const sanitizeHtml = require('../lib/sanitize-html');
+
+module.exports = function(req, res, next) {
+
+	// we don't want the lists to be out of date
+	res.set('Cache-Control', 'private, max-age=0, no-cache');
```
phamann: As discussed in person, you do not need to handle caching logic at a controller level.

phamann:
:+1:

PR #401, opened 2015-11-03
Try auto-dismissing push notifications on desktop

adgad:

Do we want to try this (Only works in Chrome 47)


laurieboyes:
Oh wow cool, did you try a fake one?

adgad:
Nope. Cowboying. (Blog post here, for reference) http://blog.chromium.org/2015/10/chrome-47-beta-idle-time-work-splash.html

laurieboyes:
:cow2: :+1:

PR #402, opened 2015-11-03
Prepare for n-quote changes (origamification)

matthew-andrews:


adgad:
Hmm, do we even use n-quote on myft-page?

matthew-andrews:
it's in the product tour i think?

keirog:
ah that part is about to die; aborting but thanks

keirog:
(See https://github.com/Financial-Times/next-myft-page/pull/404)

PR #406, opened 2015-11-03
Add to list fix

laurieboyes:
Need to handle redirecting anonymous folks to login when it's an add-content-to-list request.
Featuring myft-page's first POST smoke tests! :tada:


keirog:
:+1: with limited understanding

PR #409, opened 2015-11-04
Add push preference panel to product tour

phamann:
As discussed this adds the preference panel to enable push within the tour.

google chrome


keirog:
:+1:

PR #417, opened 2015-11-09
Remove content from list

laurieboyes:
screen shot 2015-11-06 at 15 49 31

 - Move from using next-card to templates within myft-page that are copied from next-card and modified, using (but never modifying) next-card's classes and styles
 - More complexity to lib/get-actor code (see line comment below)


server/init.js
```diff
@@ -77,8 +77,11 @@ api.get('/onsite/:userId/:relationship/:type/:subject', require('./controllers/a
 api.put('/onsite/:userId/:relationship/:type/:subject', require('./controllers/api/proxy'));
 api.delete('/onsite/:userId/:relationship/:type/:subject', require('./controllers/api/proxy'));
 
+// these need seperate endpoints as the actor is not a User
+// todo starting to think all these endpoints should have an /actorType/actorId section
 api.post('/core/list/:relationship/:type', require('./controllers/api/gateway'));
 api.get('/offsite/list/:relationship/:type', require('./controllers/api/gateway'));
+api.post('/offsite/list/:listId/:relationship/:type/:subject', require('./controllers/api/gateway'));
```
laurieboyes: Before lists, all myFT API calls had `user` as the actor type. Now `list` can be an actor (e.g. list contains content)

I've written a weird get-api-actor function to infer what the actor type and ID should be from the URL in an ad hoc way. I think it would be much more sensible to have this info clearly specified in the URL.

So e.g. the path that is currently `/onsite/`**:userId**`/:relationship/:type/:subject` would change to `/onsite/`**:actor/:actorId/:**`relationship/:type/:subject`

Does anyone have any thoughts on this?

phamann: I concur, that we should be good citizens and change the url :sadpandaface:

test/server/lib/get-api-actor.spec.js
```diff
@@ -0,0 +1,40 @@
...
+'use strict';
+
+const expect = require("chai").expect;
+const getApiActor = require('../../../server/lib/get-api-actor.js');
+
+describe('Get API actor', function () {
+
+	it('get the right actor details for various API requests', function () {
+
+		const userId = 'f0bb6f11-c034-4792-a1c0-3647825363a0';
```
phamann: Where is this from? You should use the test accounts.

phamann:
:+1:

server/init.js
laurieboyes: Agreed that we'll merge this in and tackle the tech debt as a separate piece of work

PR #419, opened 2015-11-13
Saved articles lists

laurieboyes:
Puts lists within saved articles and changes all the myFT links to point there instead.
Still behind myftLists flag
I'm working on a separate PR to DRY up the controllers for lists, saved-articles, saved-articles-list etc.

screen shot 2015-11-13 at 11 27 19


server/controllers/api/gateway.js
```diff
@@ -4,18 +4,18 @@ const apiProxy = require('../../lib/api-proxy');
 const getApiActor = require('../../lib/get-api-actor');
 const qs = require('qs');
 
-function humanReadableAction(method, relationship, subject, requestBody, actor) {
+function humanReadableActionHtml(action) {
```
laurieboyes: @keirog suggests required params should be their own params and optional ones should be in object

server/controllers/saved-articles-list.js
```diff
@@ -0,0 +1,91 @@
...
+				return article.uuid;
+			});
+
+			let viewModel = {
+				pageId: 'list',
+				noItemsMessage: ids ? 'This list is empty' : 'This list does not exist',
+				lists: lists,
+				isList: true,
+				user: {
+					id: ''
```
keirog: ?

keirog:
:+1:

PR #421, opened 2015-11-13
Dried out some of the code duplicated in the saved article lists stuff

laurieboyes:


phamann:
:+1: Looks very sabe to me. Though you lose points for your lack of ES6 one liner implicit return statements...

PR #422, opened 2015-11-13
only show lists in saved articles if you've already got some

laurieboyes:


laurieboyes:
sneaking this one in as it's miniscule and behind a flag

PR #426, opened 2015-11-17
Fix caching logic

phamann:
After recent outages we noticed that the myFT wasn't adhere to `serve-stale-if-error` rules. After investigation it seems that most myFT pages were in-fact not being cached. This is an attempt to fix that.


server/middleware/cache-control.js
```diff
@@ -15,7 +15,7 @@ function isFragment(req) {
 }
 
 function isDefaultRequest(req) {
-	return (req.originalUrl.indexOf('myft/api/onsite') > 0 && req.method.toLowerCase() === 'get');
+	return (!req.originalUrl.indexOf('myft/api/onsite') > -1 && req.method.toLowerCase() === 'get');
```
laurieboyes: String.Prototype.includes

PR #433, opened 2015-11-19
create list description field

laurieboyes:
Pretty much as you'd expect on the form:
screen shot 2015-11-19 at 16 49 27

And still pretty hideous on the lists page (but don't worry, it's only a couple of PRs away from getting more presentable):
screen shot 2015-11-19 at 16 50 29


keirog:
:+1:

PR #435, opened 2015-11-20
made lists page prettier by stealing a load of next-card styles

laurieboyes:
Does anyone have any comments on the way I've extended some of these n-card styles for the myft lists page?


keirog:
No comments :see_no_evil: :+1:

PR #436, opened 2015-11-20
Fix clippings maintenance mode

keirog:
- Fixes regex for clippings path check (the trailing slash wasn't optional)
- Allows layout override (clippings now uses vanilla layout)
- Hides the myFT nav for clippings maintenance pages

Note: isUserPage is gone as we weren't using it anywhere


laurieboyes:
zing :+1:

PR #438, opened 2015-11-20
Dashboard tweaks

keirog:
1. Increased the number of articles per topic card from 2 to 4 (5 seemed too many)
2. Fixed a flexbox alignment issue


wheresrhys:
:+1:

PR #440, opened 2015-11-20
Fixes flexbox issue in safari & firefox

keirog:


keirog:
**Safari**

Before | After
---- | ----
saf-before | saf-after

**Firefox**

Before | After
--- | ---
ff-before | ff-after

PR #445, opened 2015-11-23
Don't show opt-in/ft.com buttons on clippings pages for Next users

keirog:


laurieboyes:
:+1:

server/controllers/product-tour.js
```diff
@@ -216,6 +216,7 @@ module.exports = function (req, res) {
 		viewModel.layout = 'vanilla';
 		viewModel.clippingsLinks = getClippingsLinks(req);
 		viewModel.isLoggedOut = false;
+		viewModel.isNextUser = req.cookies.FT_SITE === 'NEXT';
```
commuterjoy: I hope cookies are removed from the backend by the CDN, so this probably will break the caching?

phamann: @commuterjoy IIRC this page is not being served on `next.ft.com` but only the interstitial `clippings.ft.com` which will not be CDN'd. The Cookies are only used to determine whether the user should bypass interstitial messaging etc.

Will check with @keirog to confirm.

keirog: @commuterjoy Yep, @phamann's right (ta)

matthew-andrews: worth adding a comment to this effect?

commuterjoy: ok, thanks.

PR #446, opened 2015-11-24
Lists page share

laurieboyes:
Pretty basic list sharing overlay functionality. Could do with a design review, but as it's behind the myftLists flag, no harm in releasing.

screen shot 2015-11-24 at 12 48 36


client/components/lists/index.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+const Overlay = require('o-overlay');
+
+module.exports = {
+	init: function () {
+
+		const shareButtons = Array.from(document.querySelectorAll('.js-share-list'));
```
phamann: Is `Array.from` safe?

server/lib/lists.js
```diff
@@ -4,6 +4,10 @@ const myFtApi = require('next-myft-client');
 
 function fetchLists(userId) {
 	return myFtApi.getAllRelationship('user', userId, 'created', 'list')
+		.then(lists => {
+			lists.items.forEach(list => list.publicUrl = 'https://next.ft.com/myft/list/' + list.uuid);
```
phamann: Potentially use a `map` instead?

client/components/lists/index.js
laurieboyes: Works for me locally and looks good on this table: http://kangax.github.io/compat-table/es6/#Array.from

phamann:
:trollface: Would be nice to see `o-share` instead here. But thats for another day :wink:

phamann:
:+1:

PR #448, opened 2015-11-26
put in some api routes still required for apps using myft-client pre v6.0.0

laurieboyes:
Whoops! That'll teach me to do all my manual testing myFT button testing within the myFT app.
Patch for the hole in the tests coming later.


phamann:
SHIP IT!

PR #451, opened 2015-11-26
fixed and tested add to list offsite route

laurieboyes:


laurieboyes:
IRL :+1: from @phamann

PR #454, opened 2015-11-27
add to new list enhanced

laurieboyes:
Hmm, what do you think of my approach to the 'no name' error message. What I like about it is that it gives us a way to style multiple elements of a form group to react to a specific error, just by adding classes to the form group element. This is nice and convenient both server-side (core) and with JS


phamann:
:+1: Agree that the error message solution is well suited.

PR #455, opened 2015-11-30
removed portfolio overview from myFT

laurieboyes:
This includes removing handlebars from client, which, arguably, I should never have brought in in the first place.


keirog:
:+1:

PR #456, opened 2015-11-30
fixed add-to-list error styles on other apps

laurieboyes:
Should probably do this: https://github.com/Financial-Times/next-myft-ui/issues/67 but this will solve the current problems caused by this for now.


keirog:
:+1:

laurieboyes:
ty

PR #457, opened 2015-11-30
Refactor auth

phamann:
Looks big but actually doesn't introduce much new code. 

Simply abstracts each authentication strategy into it's own middleware, with the key difference now being an auth middleware must decorate `req.authenticated = true` any subsequent auth middleware will then be skipped.

The middle order is now declared and managed in the main routes file, so we can be more explicit of intent. Separating like this gives us the added benefit to be able to use different auth ordering for eahc sub router (such as auth, api & default).

### Todo:
- [ ] Refactor controllers to use new `req.authentication` object
- [ ] Add auth middleware for public lists


laurieboyes:
I've made a PR into this one: https://github.com/Financial-Times/next-myft-page/pull/458. Also some of the tests need a look at.

Apart from that, woop woop! Lovely and clean

phamann:
:ship:'ing

PR #458, opened 2015-12-01
Refactor auth portfolio purge

laurieboyes:
I removed the portfolio code the other day as it's unused and will soon be replaced and the flag was expiring.


keirog:
:+1:

PR #461, opened 2015-12-02
Fix caching logic :/

phamann:
It turns out express middleware gets executed in order of routers, i.e. all global middleware mounted on `app` gets executed before ones mounted on individual instances of `express.Router()`. Not in sequential source order within the routes file, like I thought :(

I've also added a test to ensure this doesn't happen again.


laurieboyes:
:+1:

keirog:
:+1:

PR #463, opened 2015-12-02
Fix false positives for `isRss` and `isJson`

keirog:
`/myft` wasn't always in the path, causing some false positives


keirog:
Got IRL thumbs up from @phamann

PR #465, opened 2015-12-02
Copy to list enhanced

laurieboyes:
Take advantage of myft-ui's new copy-from-list-to-list overlay. Also a bit of copy fiddling and a small bit of tidying


laurieboyes:
IRL :+1: from @keirog

PR #471, opened 2015-12-03
Add UI to toggle list privacy

phamann:
public-private

- Bumps client to support `updateActor`, `purgeActor` and data sanitization
- Add toggle form to lists page
- Add `/core/:actor/:actorId` route to enable manipulation of nodes
- Default lists to private


views/lists.html
```diff
@@ -43,9 +42,22 @@ <h3 class="n-card__headline">
...
+								<form class="myft-lists__item__action myft-ui myft-ui--prefer" method="POST"
+									action="/__myft/api/core/list/{{uuid}}?method=put">
+									<input type="hidden" name="isPublic" value="{{#if isPublic}}false{{else}}true{{/if}}" />
+									<button
+										type="submit"
+										class="myft-ui__button{{#variant}} myft-ui__button--{{this}}{{/variant}}"
+										data-trackable="set-public"
+										{{#if isPublic}}aria-pressed="true"{{/if}}
+										><span class="myft-ui__core-text">{{#if isPublic}}Yes{{else}}No{{/if}}</span></button>
+									<label>Public:</label>
```
keirog: Not sold on the <label> to nowhere. Also could it be above the button?

keirog:
Other than the label comment, pragmatic thumb :+1:*

*We need to revisit how we do our togglers at some point™

keirog:
:ship:

PR #474, opened 2015-12-03
Create list enhanced

laurieboyes:
Got rid of the form at the bottom of the lists page and replaced it with this

screen shot 2015-12-03 at 17 15 37

screen shot 2015-12-03 at 17 15 56


keirog:
:+1:

PR #478, opened 2015-12-04
Ensure user is logged in before checking list privacy

phamann:
ronseal


keirog:
Got a failing test but otherwise :+1:

PR #479, opened 2015-12-04
refresh page after list created so you can see it

laurieboyes:
Goes with this: https://github.com/Financial-Times/next-myft-ui/pull/79


keirog:
MVPlus1? :+1:

PR #484, opened 2015-12-07
Make maintenance message semi-configurable

keirog:
Example usage:
`maintenance(req, res, 'Feature', { textHtml: 'new message' }`


laurieboyes:
:+1:

PR #486, opened 2015-12-07
Fix list page cache logic

phamann:
Our default `cache-control` middleware runs before the custom middleware used to authenticated list pages, this results in a surrogate key of `undefined:page`. To fix this I have just manually set the header to the lists ID, which in-fact is correct rather than the users id.


keirog:
:+1:

PR #488, opened 2015-12-07
Add new controller to handle list updates

phamann:
This adds a new route `/list/update` to handle updating of list data and adding content to lists. Sadly in cannot be truly RESTful via a `PUT /list`  as HTML forms do not allow for a PUT method so have settled with its on route. :sadpandaface:

The main reason for doing this was to handle empty submissions of add content to list form, but has the added benefit that doing the work to edit lists should now be trivial.


server/controllers/lists/create.js
```diff
@@ -13,7 +13,7 @@ module.exports = (req, res, next) => {
 		req.body.contentUuid = sanitizeHtml(req.body.contentUuid);
 	}
 
-	if(!req.body.name) {
+	if(!req.body.name || req.body.name === '') {
```
keirog: But `!!'' === false` anyway, no?

server/controllers/lists/update.js
```diff
@@ -0,0 +1,62 @@
...
+
+const apiProxy = require('../../lib/api-proxy');
+const sanitizeHtml = require('../../lib/sanitize-html');
+
+module.exports = (req, res, next) => {
+	let contentUuid;
+	let listId;
+	req.body.name = sanitizeHtml(req.body.name);
+
+	if(!req.body.uuid) {
```
laurieboyes: These core experience error redirects are stale and possibly a bit rubbish anyway, I think Keith might have been looking at this for another controller

server/controllers/lists/update.js
```diff
@@ -0,0 +1,62 @@
...
+	apiProxy({
+		actorName: 'list',
+		actorId: listId,
+		type: 'list',
+		method: 'put',
+		body: req.body
+	}, res.locals.flags)
+	.then(() => {
+
+		console.log('Updated list');
```
laurieboyes: accidental log?

views/partials/list-form.html
```diff
@@ -6,16 +6,17 @@
...
 
 		<form class="myft-list-form" method="POST"
-			  action="/__myft/api/core/list/contained/content?method=post">
-			<input type="hidden" name="uuid" value="{{contentId}}"/>
-				<div class="o-forms-group">
-					<select name="subject" class="o-forms-select js-list-select">
-						<option>Choose a list</option>
+			  action="/__myft/list/update">
+				<input type="hidden" name="contentUuid" value="{{contentId}}"/>
+				<div class="o-forms-group {{#ifEquals error 'noListId'}}o-forms--error n-myft-ui__error--no-name{{/ifEquals}}">
```
laurieboyes: I guess as it's currently the only error on the form it doesn't matter that it's got a weird name but it still bothers me a little bit. I'm really not sure why I put these error styles in myft-ui (https://github.com/Financial-Times/next-myft-ui/blob/master/main.scss#L176-L183). Maybe move them back into myft-page?

phamann: They are still needed in `myft-ui` as it is loaded in non myFT pages such as the homepage and articles.

phamann:
Dependent on: https://github.com/Financial-Times/next-myft-ui/pull/81

laurieboyes:
:+1:

PR #490, opened 2015-12-08
Refactor list card view sorting logic

phamann:
- Zip the myFT relationship into the content object
- Use list creation date in sort
- Replace content timestamp with relationship created in sort


laurieboyes:
:+1:

PR #495, opened 2015-12-08
Tidy up list action buttons on < S layouts

keirog:
Looks marginally less broken.

Before:
before
After:
after


laurieboyes:
Ha, nice :+1:

PR #496, opened 2015-12-09
Break up long words for list-title headings

keirog:
Avoided applying this to all *my*FT headings, in case of dodgy topic
word-breaks.


laurieboyes:
:+1:

PR #498, opened 2015-12-09
Fix flash messaging for core experience

phamann:
#### Before
screen_shot_2015-12-08_at_13 39 30

#### After
google chrome


laurieboyes:
:+1: niiice

PR #500, opened 2015-12-09
Break all long-worded headings

keirog:
Realised all the headings are non-article headlines, so it’s safe to break words


laurieboyes:
:+1:

PR #502, opened 2015-12-09
temporarily don't do a purge after list delete, while we've rolled back myft-api for debugging purposes

laurieboyes:
temporarily don't do a purge after list delete, while we've rolled back myft-api for debugging purposes


keirog:
:+1:

PR #506, opened 2015-12-10
Cleanup unused read list code

phamann:
- Authentication is now handled by the middleware
- We no longer have lists navigation


laurieboyes:
:+1:

PR #510, opened 2015-12-10
Add labels to the list forms

keirog:
Ronseal.

**~~~ Small screens ~~~**

Before | After
--- | ---
small-before | small-after

**~~~ Big screens ~~~**

Before | After
--- | ---
big-before | big-after


keirog:
Got an IRL :+1: from @pauloneillft

PR #511, opened 2015-12-10
remove legacy list routes and controllers

laurieboyes:
And updated client readme https://github.com/Financial-Times/next-myft-client/pull/75


keirog:
:+1:

PR #513, opened 2015-12-10
Update clippings launch message

keirog:


keirog:
Got an IRL :+1: from Max

PR #516, opened 2015-12-14
Refactor saved-articles and lists to use v3 ES index

phamann:
This is the first major PR to refactor myft-page to use the more stable v3 ES index. 

- Create new `get-content` lib
    - Use signed fetch
    - Only request subset of fields from content model
- Create `v3-to-card' model to interface the new data to the model required by `next-card`
- Refactor `readAll` lists controller to make a single request to ES rather than one per list.
- Refactor saved-articles to use `get-content`
- Refactor read list to use `get-content`
- Bump the `article-primary-tag` lib to v2.0.0


phamann:
I hoping this should have a substantial improvement on the performance of lists pages.

phamann:
cc @matthew-andrews

server/controllers/lists/readAll.js
```diff
@@ -1,28 +1,39 @@
...
+		return res;
+	}, {});
+};
+
+const hydrateListContent = (rawLists) => {
+	const listPromises = pluck('uuid', rawLists).map(getList);
+	let lists;
+
+	return Promise.all(listPromises)
+			.then(tap(listData => lists = listData))
```
laurieboyes: ahhhhh took to a while to work out what this does but it's very clever and much nicer than passing ever growing objects along promise chains. nice.

server/controllers/saved-articles.js
```diff
@@ -48,9 +50,10 @@ module.exports = function(req, res, next) {
 				viewModel.content = [];
 				render(viewModel);
 			} else {
-				return getLegacyContent(content.items, res.locals.flags).then(content => {
-					viewModel.content = content;
-					viewModel.content.forEach(a => a.isSaved = true);
+				return getContent(content.items.map(i => i.uuid)).then(content => {
+					viewModel.content = content
+											.map(v3toCard)
+											.map(tap(c => c.isSaved = true));
```
laurieboyes: although I don't understand the usage of tap here?

laurieboyes:
:+1:

PR #517, opened 2015-12-15
Pagination

phamann:
In response to user facing issues for customers with large amounts of clipped items, this adds pagination to the saved-articles and list view pages. It also includes some _Quick Fixes™_ to make the `readAll` lists controller more efficient.

- Add per page lib
- Refactor `fetch-list` and `fetch-list-content`
- All saved-articles and read list to accept `page` query param
- Pagination view partial
- Quick styling using o-buttons

### Screenshot
google chrome

### Todo
- [ ] Expose a `raw` parameter when querying a `node` via the myFT API to ensure relationships are not returned for lighter payload.
- [ ] pagination UX


matthew-andrews:
:+1:

PR #518, opened 2015-12-16
core experience route for remove content from list

laurieboyes:
Fixes https://trello.com/c/PuFcy5FB/220-core-experience-articles-are-not-being-removed-from-lists


keirog:
:+1: :santa:

PR #519, opened 2015-12-16
remove select list placeholder now that we have a label

laurieboyes:
Demands an unnecessary extra click. Default option is most recently created list


keirog:
:+1:

PR #522, opened 2015-12-16
Change from register to subscribe on product tour link bar

keirog:


laurieboyes:
:+1:

keirog:
Many thanks to @laurieboyes for pointing out a stupid typo

PR #523, opened 2015-12-17
Refactor list fetching

phamann:
Currently to construct the `/myft/lists` page we are making a DB HTTP request per list and not limiting the amount of lists we fetch. For users with a lot of lists and list content this is a very expensive operation and has potential to fail. Using the new [custom endpoint](https://github.com/Financial-Times/next-myft-api/pull/109) exposed on the myft-api we can wrap this up into a single call which supports pagination.

This PR refactors all locations in the code base that we fetch all lists for a user, and introduces pagination functionality to the all lists page.


server/lib/fetch-lists.js
```diff
@@ -0,0 +1,46 @@
...
+		credentials: 'include',
+		headers: {
+			'Content-Type': 'application/json',
+			'X-API-KEY': process.env.USER_PREFS_API_KEY || process.env.MYFT_API_KEY
+		}
+	};
+
+	let query = '';
+
+	Object.keys(data || {}).forEach(function(key) {
```
keirog: =>

keirog:
Niiice :+1:

PR #526, opened 2015-12-17
stop hard-coding each core path in gateway login redirect

laurieboyes:
especially as one of them was broken due to not being updated in sync with the routing


keirog:
:+1:

PR #528, opened 2015-12-17
Add default messaging for users with no lists

phamann:
google chrome


laurieboyes:
:+1: except has James signed off on the copy?

phamann:
Copy has been updated

PR #531, opened 2015-12-18
redirect to myft page after login rather than always product tour

laurieboyes:
Redirect will go to product tour if you're still not following anything, or feed/dashboard (depending on preference) if you are


laurieboyes:
Ahh wait it's just occurred to me that this fixes this link
screen shot 2015-12-18 at 09 47 30
but not this one 
screen shot 2015-12-18 at 09 47 42

laurieboyes:
Had a brief look at dropping a hack into product tour to redirect based on the referrer for a more general fix but of course you don't get a referrer from https. Still worth merging?

keirog:
:+1:

PR #540, opened 2016-01-04
Move catch logic in list auth to prevent runtime error

phamann:
The location of this `catch` block was causing the proceeding `then` to still be executed even if the response has already been sent. This was causing a runtime error for all list 404s. 

Fixes this error in Sentry: https://app.getsentry.com/nextftcom/ft-next-myft-page/issues/100900825/events/3921911032/


laurieboyes:
:+1:

PR #544, opened 2016-01-06
No follow but saved journey

laurieboyes:
https://trello.com/c/e8K7ttIF/278-change-routing-logic-for-clippings-migrants-so-that-if-you-re-not-following-anything-but-have-saved-some-articles-you-can-easily

Added in some extra stuff for if you have lists but no saved articles (which is highly unlikely unless you're deliberately trying to break it but I think still worth accounting for)

Logic summarised in root-redirect test names:
* it should redirect to my news if user follows topics
* should redirect to my news if user follows topics regardless of anything else you've got
* should redirect to saved articles when user follows no topics but has saved articles
* should redirect to saved articles when user follows no topics but has saved articles, whether or not user has lists
* should redirect to saved articles list view when user has no topics and no saved articles but has non-redirect lists
* should redirect product tour when user has no follows or saved or lists
* should redirect product tour when user has no follows or saved and only redirect lists


server/controllers/root-redirect.js
```diff
@@ -0,0 +1,49 @@
...
+	if (err.message === 'No user data exists') {
+		return false;
+	} else {
+		throw err;
+	}
+}
+
+module.exports = function (req, res, next) {
+
+	// todo bit worried about this and caching
```
laurieboyes: Any thoughts on this?

server/controllers/root-redirect.js
```diff
@@ -0,0 +1,49 @@
...
+	} else {
+		throw err;
+	}
+}
+
+module.exports = function (req, res) {
+
+	// todo bit worried about this and caching
+	//  - these api requests obviously won't be cached
+	//  - the myft/ page will be cached?? need to invalidate when something changes?
```
laurieboyes: any thoughts on *this*

keirog:
:+1: Plus offline comments about API calls

laurieboyes:
@keirog had a good idea that rather than making three separate calls here we could perhaps make a custom route in myft-api for getting a summary of what the user has (e.g. following 0, saved 4, lists 2) and just make the one request.

Considering merging this in for now, as it fixes the bug, and making this optimisation later

PR #545, opened 2016-01-07
product tour get started to work in core and enhanced

laurieboyes:
Moved logic for product tour get started button redirect from client to server, so that it works on core.

Required poking a hole in session auth middleware to allow controller to be available for anonymous users, while still providing it with userId if logged in.

Also minor style fix on product tour links and buttons.

https://trello.com/c/GV8FQPT3/257-core-experience-get-started-button-does-not-work-on-the-product-tour-page

Todo:
- [ ] write some tests


server/controllers/product-tour-get-started.js
```diff
@@ -0,0 +1,29 @@
...
+		return false;
+	} else {
+		throw err;
+	}
+}
+
+module.exports = function (req, res) {
+
+	if(!res.locals.userId) {
+		res.redirect(302, '/products?segID=400863&segmentID=190b4443-dc03-bd53-e79b-b4b6fbd04e64');
```
keirog: return

keirog:
Other than commented, :+1:

PR #554, opened 2016-01-12
Fix BB10 list toggler layout

keirog:


keirog:
Before, on BB10: Label + button = two lines
After, on BB10: Label + button = one line

laurieboyes:
:+1:

PR #555, opened 2016-01-13
Add users followed topics to outbound content links

phamann:
In conjunction with [#1206 in next-article](https://github.com/Financial-Times/next-article/pull/1206) this ensures that a users followed topics are visible when viewing content pages. This is acheived by adding a query param of `?myftTopcis=` to all outbound content links from my-ft pages.

- Refactor referrerTracking to append topic ids
- Add tests


phamann:
The tests are failing due to a reliance on: https://github.com/Financial-Times/n-topic/pull/30

laurieboyes:
:+1:

PR #557, opened 2016-01-13
Only apply topics to links on myFT news feed.

phamann:
Ensure that `content.primaryTag` is an `Array` before applying to `referrerTracking`


laurieboyes:
:+1:

PR #559, opened 2016-01-19
Split out gateway messages

keirog:
We'll be using them outside of the generic gateway controller (for the follow-email work)


laurieboyes:
:+1: noice

phamann:
:+1: FYI I actually want to refactor this entire flow in the near future, as we should never be returning a response from `POST`'s requests. And instead something along the lines of the flash messaging system we created for lists.

keirog:
Ok great, was wondering about gateway's non-PRG-ness earlier

PR #561, opened 2016-01-20
Remove stickyFooter property

keirog:
No longer in use since [n-layout v6.1.0](https://github.com/Financial-Times/n-layout/releases/v6.1.0)


laurieboyes:
:+1:

PR #563, opened 2016-01-21
remove portfolio stuff from tour

laurieboyes:
Preserved our linebreak neatness hack:

screen shot 2016-01-21 at 09 46 51
screen shot 2016-01-21 at 09 47 02
screen shot 2016-01-21 at 09 47 13


keirog:
:+1:

PR #565, opened 2016-01-25
Add subscribedToDaily prop to the follow model

keirog:
This'll be used to check whether or not to display the "Topic added to
daily digest email" message for core users. (The template lives in
next-myft-ui.)


phamann:
:+1:

PR #566, opened 2016-01-25
Refactor layout preferences and render-stream

phamann:
This separates out the `get-layout-preference` logic from the `render-stream` library so that we can decide a users layout preference earlier within the controller logic of each view. Whilst it adds more complexity to each controller it simplifies the responsibility of `render-stream`. But the main reason I am doing this is to enable us to track users feature usage, which is layout specific.


laurieboyes:
:+1:

keirog:
:+1:

PR #573, opened 2016-02-02
Update the myFT header

keirog:
**Before (default layout):**
before-default

**Before (S+ layouts):**
before-s-plus

**After (default layout):**
after-default

**After (S layout):**
after-s

**After (M+ layouts):**
after-m-plus

**Rollovers:**
after-rollovers

(Contrast checks all pass)


phamann:
:+1: Looks good to me. Where are we at with the :eggplant: colour discussion?

keirog:
Ah sorry, I took the woohoo as a :+1:!

The :eggplant: discussion is still not settled and seems quite a way off still, from what I can gather. Clem is happy with Origami's blue-1 for now.

PR #574, opened 2016-02-02
Update header colour

keirog:
after

It was decided that blue-1 wasn't an appropriate interim color (before
we have an official brand color), so we're going with this for the time
being.


phamann:
:+1: I still do understand why we have "interim" colours? What is blocking Clem from deciding a brand colour?

keirog:
The proposed aubergine doesn't look good with claret, so we're not ready to use it yet. Plus we're trying not to diverge from o-colors in the meantime.

PR #578, opened 2016-02-04
Use core/enhanced classes instead of no-js/js

keirog:
This is part of a site-wide transition to use the former over
the latter, in order to better reflect our usage.


laurieboyes:
Just wonderful :+1:

PR #580, opened 2016-02-05
Add "Edit topics" to the Following page

keirog:
Getting something out there so we can all play with it (behind the `myFtEditTopics` flag).

**Still do to:**
- [ ] Decide whether or not this is the right solution. If so:
- [ ] Discuss the toggle (up/down) arrow as a common icon & implement
- [ ] Resolve issue of loadsa follows
- [ ] Tidy up styling

Brace yourself for…
**Screenshots at different layouts**

Default|M|XL
---|---|---
||


client/components/edit-topics/index.js
```diff
@@ -0,0 +1,15 @@
+const expander = require('o-expander');
```
laurieboyes: hipster potential: User `import` rather than `require`

views/partials/toolbars/edit-topics.html
```diff
@@ -0,0 +1,13 @@
+{{#if followed}}
+<div class="edit-topics js-edit-topics o-expander" data-o-component="o-expander" data-trackable="edit-topics">
+    <button class="edit-topics__toggle o-expander__toggle ijfsyyyyyy">Edit topics</button>
```
laurieboyes: is the `ijfsyyyyyy` class intentional?

laurieboyes:
:+1: and comments

views/partials/toolbars/edit-topics.html
keirog: Oh crikey, forgot about that… thanks

client/components/edit-topics/index.js
keirog: Yessss

PR #587, opened 2016-02-10
Fix layout preference routing

phamann:
/cc @keirog


server/lib/get-layout-preference.js
```diff
@@ -1,13 +1,13 @@
...
 
 module.exports = function getLayoutPreference(userId) {
 	return myFtApi
 			.getRelationship('user', userId, 'preferred', 'preference', 'myft-layout')
 			.then(res => {
-				return layouts.get(res._rel.layout || 'default');
+				return res._rel.layout || 'topic';
 			})
 			.catch(() => {
-				return layouts.get('default');
```
keirog: one-liner police but wevs

keirog:
:+1:

PR #588, opened 2016-02-10
Edit-topics tweaks

keirog:
Resolves:
- visual alignment on small screens
- follow buttons hitting the container border
- seeing the follow buttons flash up on page load
- core experience looking shoddy

Before (mobile):|After (mobile)
---|---
before-m|after-m

**After (core):**
after-core
^ That's just a quick fix until we resolve the issue of main headings on all our pages


laurieboyes:
:+1:

PR #591, opened 2016-02-11
Make articles/topic cards fade out if their respective topics are unfollowed

keirog:
**Time:**
unfollow-fade-time

**Topic:**
unfollow-fade-topic


phamann:
:+1: Nice!

PR #604, opened 2016-02-17
Swap app and init filenames and references

keirog:


laurieboyes:
:+1:

package.json
```diff
@@ -10,7 +10,7 @@
   ],
   "scripts": {
     "test": "make test",
-    "start": "./node_modules/.bin/forever server/app.js"
+    "start": "./node_modules/.bin/forever server/init.js"
```
matthew-andrews: `./node_modules/.bin` shouldn't be needed inside `package.json` scrypts

Dockerfile
```diff
@@ -1 +0,0 @@
-FROM financialtimes/next-heroku:0.12.6
```
matthew-andrews: lol wow

package.json
keirog: I suppose we shouldn't encourage using `npm start` anyway. Will remove

matthew-andrews: Haikro will happily fall back to using npm start if no Procfile.

Also you shouldn't need `./node_modules/.bin` inside the Procfile either

keirog: Ok ta

PR #605, opened 2016-02-17
Path has-bin (sorry)

keirog:


phamann:
:+1: Based on recent findings we would even consider ditching `forever` entirely?

phamann:
Heroku Dyno crash restart policy: https://devcenter.heroku.com/articles/dynos#dyno-crash-restart-policy

keirog:
"After the 320 minute cool-off period is reached, restart attempts happen every 320 minutes"

keirog:
I'll merge this in and we can decide about forever later.

Forever later™

PR #608, opened 2016-02-23
New Explore page skeleton, with React/GraphQL spangles

keirog:
- Copies React rendering/GraphQL polling functionality from [next-front-page](https://github.com/Financial-Times/next-front-page)
- Introduces the new Explore page (behind the `myFtExploreReact` flag), with just popular topics for now

**Note:**
- front-page’s section & card components were abstracted out to be [n-section](https://github.com/Financial-Times/n-section) and [n-card](https://github.com/Financial-Times/n-card), respectively
- no client-side React bits added yet
- umm… no tests yet


config/layouts/index.js
```diff
@@ -0,0 +1,5 @@
+import stripe from './stripe';
```
phamann: I've been calling this "stripe" but think we should follow the homepage's convention of "section". Thoughts?

config/layouts/stripe.js
```diff
@@ -0,0 +1,18 @@
+import { Column } from '@financial-times/n-section';
+import { Row } from '@financial-times/n-section';
+import { Content } from '@financial-times/n-section';
+
+export default [
+	{
+		type: Row,
+		components: Array(4).fill().map(() => {
```
phamann: `Array(4).fill()` :bomb:

package.json
```diff
@@ -14,6 +14,14 @@
   },
   "dependencies": {
     "array.prototype.find": "^1.0.0",
+    "babel-core": "^6.5.2",
+    "babel-plugin-add-module-exports": "^0.1.2",
+    "babel-plugin-array-includes": "^2.0.3",
+    "babel-plugin-transform-es2015-destructuring": "^6.5.0",
```
phamann: Might not need this as and `includes` above as we're rocking Node v5 now

config/layouts/index.js
keirog: Yep makes sense; think I was too zoomed in at that point to realize

server/app.js
```diff
@@ -3,15 +3,30 @@
...
+import ReactServer from 'react-dom/server';
+import { start as startPolling } from './libs/graphql-poller';
+// TODO: put n-card & n-section helpers in n-utils?
+import { colspan } from '@financial-times/n-section/libs/helpers';
+
 // Create the app
 const app = module.exports = express({
 	helpers: {
-		'ifThird': function(i, o) { return (i === 2) ? o.fn(this) : o.inverse(this); }
+		reactRenderToString: (componentClass, props) => {
```
phamann: Whilst I know the homepage had this convention, would be good to abstract this into its own module file to keep the `app.js` file lean.

server/init.js
```diff
@@ -1,5 +1,6 @@
-/*eslint new-cap:0*/
 'use strict';
+
+require('./setup');
 const app = require('./app');
 const logger = require('ft-next-express').logger;
```
phamann: Why is logger here?

laurieboyes:
hip :+1:

server/libs/graphql-poller.js
```diff
@@ -0,0 +1,66 @@
...
+			});
+		}
+		return readyPromise;
+	},
+	getData: function (type) {
+		return queryResults[type].data;
+	},
+	getLastFetchedTime: function (type) {
+		return queryResults[type].lastUpdated;
+	}
```
phamann: hipster missed one-liner opportunity police on the above two methods.

package.json
keirog: Nah I've tested without these—Node v5 isn’t as hipster as we want yet :cry:

phamann:
:+1: :palm_tree: :100: Nice work!

server/init.js
keirog: Because of `logger.info` on [line 15](https://github.com/Financial-Times/next-myft-page/blob/9142809347b603ed923d36e37a0d7cc8fade53e5/server/init.js#L15). Seems useful to have the port logged out still, no?

PR #611, opened 2016-02-23
Use n-logger directly instead of express instance

phamann:
ronseal


package.json
```diff
@@ -40,7 +42,6 @@
     "handlebars": "^3.0.3",
     "md5": "^2.0.0",
     "n-eager-fetch": "^1.0.3",
-    "@financial-times/n-section": "^1.0.0",
```
keirog: Ooh ta, forgot to reorder

keirog:
:+1:

PR #614, opened 2016-02-23
Explore search

laurieboyes:
screen shot 2016-02-23 at 17 30 58


client/components/explore/main.scss
```diff
@@ -0,0 +1,64 @@
...
+
+.topic-search__intro {
+	@include oTypographySans(s);
+	color: oColorsGetPaletteColor(cold-1);
+	text-align: center;
+}
+
+.topic-search__input {
+	@include oFormsCommonField();
+	@include oTypographySansData(m);
```
laurieboyes: We shouldn't _really_ be overriding origami component fonts, but Paul was very grateful

keirog: I'm not keen on this… Why override just this one and not other form styles?

keirog: Can we check with Alberto about changing it for all o-forms controls?

keirog: As per our offline chat, I'm backing down on this :see_no_evil:

keirog:
:+1:

phamann:
:v:

PR #616, opened 2016-02-24
hide topic search in core experience

laurieboyes:
Nicer than a dead input field:

screen shot 2016-02-24 at 14 35 36


keirog:
Ideally it'd work in core too, but :+1: for now™

PR #623, opened 2016-02-26
Sanitize explore topics

phamann:
This sanitizes the sections in the explore page to remove topics that a user is already following and then de-duplicate topics across the sections, hopefully resulting in every topic being unique. 

To achieve this we have to query for more than 4 topics per section to allow a buffer if a user is already following items in the first 4 topics. To start with I have set this value to 12.

Currently, If each process produces less than 4 topics, we back fill the collection with excess topics, which therefore, could result it us displaying un-wanted topics on rare occasions. We'll ship this first to see how much that situation occurs. I've discussed with James other data sources where we could get the excess topics from.

#### Todo:
- [x] Remove follows 
- [x] Dedupe topics
- [x] TESTS!


phamann:
Will not ship until I have written some tests for the `lib/sanitize-explore-topics.js` lib

server/controllers/explore.js
```diff
@@ -2,28 +2,37 @@ import { Section } from '@financial-times/n-section';
...
-	Promise.all(sectionPromises)
-		.then(([fromYourIndustry, recommendedTopics]) => {
-			const data = {
-				fromYourIndustry,
-				recommendedTopics,
-				todaysTopics: getData('todaysTopics'),
-				popularTopics: getData('popularTopics')
-			};
+	Promise.all(dataPromises)
+		.then(([follows, ...data]) => sanitizeTopics(data, follows))
```
keirog: Niiiiice

keirog:
:+1: :sparkles:

PR #627, opened 2016-03-01
handle case where your only saved articles can no longer be found in es

laurieboyes:
As was the case with the next premium test user


phamann:
:+1:

PR #630, opened 2016-03-03
Stop rogue-including fonts

keirog:
- Bumps o-expander to v3.0.0-beta.1, as v2 was calling `oFontsInclude` which was not cool
- Adds a temporary™ bower resolution while we're in o-header flux
- Sorts bower deps

Unfortunately we can't yet bump to the stable v3.0.0 of o-expander, as that introduces a dependency on the latest, (non-next-branch) o-ft-icons. (See [o-expander releases](https://github.com/Financial-Times/o-expander/releases).)

Expanders tested and still working fine.


laurieboyes:
:+1:

PR #631, opened 2016-03-04
Product tour reup

laurieboyes:
Replaced the standfirst and follow button demo images as per the reup: https://trello.com/c/MDFGEKkl/35-product-tour-assets-re-up

Corrected some deeply misguided decisions about how we should handle responsive images across the product tour (my decisions)

Removed some dead stuff

screen shot 2016-03-07 at 12 11 59


keirog:
The little positioning nudges make me feel a bit iffy but no biggie :+1:

PR #632, opened 2016-03-07
Explore page resilience

phamann:
Currently id one of the fetch requests for recommendation content fails, it prevents the entire page from rendering. 

This simply ensures each of the libraries returns and empty data set to allow the promise chain to continue. This will result in that specific section from being rendered, but at least lets user see the page rather than a timeout.


server/lib/fetch-recommended-topics.js
```diff
@@ -1,5 +1,12 @@
+import logger from '@financial-times/n-logger';
 import fetchGraphQlData from './fetch-graphql-data';
 import { recommendedTopicsByUserId as recommendedByUserQuery } from '../../config/queries';
 
 export default (userId, flags) => fetchGraphQlData(recommendedByUserQuery(userId), flags)
-									.then(data => data.user);
+									.then(data => data.user)
```
keirog: Weird spacing?

laurieboyes:
:+1:

keirog:
:+1:

PR #633, opened 2016-03-07
Add a fllback industry if user doesn't have one defined

phamann:
In turns out some users don't have `industry` defined in their membership session, this was resulting in an error being thrown from the HUI api. A temporary solution is to fallback to the most popular industry: "Financial services". We are also logging these occurrences to gauge how often they occur.


keirog:
:+1:

PR #635, opened 2016-03-08
Distinguish between Explore link clicks

keirog:
(Previously had the same tracking domPath as the main Explore nav link)


phamann:
:+1:

PR #640, opened 2016-03-10
Add sanitization logic to AB test recommends data

phamann:
Forgot to add this is in, which was resulting in people being suggested topics they already follow :/


keirog:
:+1:

PR #642, opened 2016-03-10
Update partial state card

keirog:
- only shows 1 partial card if user is following under 4 topics
- amends copy to encourage users to follow at least 4

**Time view**

Before | After
---|---
before-time | after-time

**Topic view**

Before | After
--- | ---
before-topic | after-topic


client/components/dashboard/main.scss
```diff
@@ -40,21 +40,29 @@
 }
 
 .myft__dashboard-card--empty {
+	@include oGridRespondTo($until: L) {
+		min-width: 100%;
+	}
+
 	.n-topic {
+		align-items: center;
```
phamann: :heart_eyes: What is this dark magic flex box foo?

server/controllers/following/layout.js
```diff
@@ -91,7 +91,7 @@ module.exports = (req, res, next) => {
 				referrerTracking: '#myft:my-news',
 				followed,
 				showEmpty,
-				emptySize: Array(showEmpty ? (4 - followed.length) : 0).fill(1),
+				emptySize: [parseInt(showEmpty, 10)],
```
phamann: Nice!

phamann:
:+1:

client/components/dashboard/main.scss
keirog: A lot of flexbox seems mad, but this one I like :smile:

keirog:
(Am ignoring the discussion about the plus symbol btw, in order to actually ship the thing)

PR #643, opened 2016-03-10
See ya mypage

keirog:
Removes the now-redundant `mypage` tracking string.

Thought we may as well remove it now, seeing as we've just changed all
the tracking domPaths anyway


laurieboyes:
:+1: good thinking

PR #644, opened 2016-03-14
Add temporary route for recommendation email feedback

phamann:
Hacky. To gather qualitative feedback, we are going to email 2500 users to see if they found the  recommendations useful. Therefore we needed a way of forcing them into the `myft` variant of the test.


laurieboyes:
:+1:

PR #645, opened 2016-03-16
use zIndex mixin for notification container

andygnewman:
cc: @keirog 

whilst on bug challenge, tidying up some of our existing z-index declarations in favour of the mixin with use cases.


keirog:
:+1::+1::+1:

PR #646, opened 2016-03-17
Fix section title clash in Explore

keirog:
The n-section styling has changed, which broke our Explore section
overrides. I've checked with Clem and she's happy for us to use the
default n-section title styling. (We're still making some minor spacing
tweaks, but small moves…)

**Before:**
before

**After:**
after


laurieboyes:
:+1:

PR #657, opened 2016-03-23
Reduce blocking request chain for following controllers

phamann:
I didn't think this through when refactoring the controllers last month. Currently we are making a blocking call to fetch the users followed topics to decide whether we should `redirectOrReturn` before then fetching the needed content if to render the controller. This has the obvious side-effect of slowing down the response time of the view by a whole round trip.

This refactors each controller to fetch both sets of data con-currently to reduce response time.


keirog:
:+1:

PR #658, opened 2016-03-24
Use n-card concept cards instead of n-topic

keirog:
**Before:**
before

**After**
after

**To do:**
- [x] Add dividing line for recommendation slice


laurieboyes:
:+1:

PR #659, opened 2016-03-29
Fix "change industry" link style

keirog:
Removed a leftover `display: block` that was only there for the previous markup.

Before:
before

After:
after


phamann:
:+1:

PR #662, opened 2016-03-29
Refactor topic-card lib to retrun empty topic cards

phamann:
Previously,  topics which had no content in the last 7 days would get filtered out of the topic-card library as it simply groups the content by topic. This refactors the library to also use the users followed topics and generates empty topic cards, so than an empty state message can be shown.

google chrome

This should hopefully reduce confusion as to why your followed topics don't appear.

Depends on [#23](https://github.com/Financial-Times/n-card/pull/23) of n-card.


phamann:
Added tests :)

keirog:
:+1:

PR #670, opened 2016-03-31
Refactor Makefile to use n-makefile and port client-side build

phamann:
To enable us to compile React components such as `n-section` & `n-card` on the client we need to able to have custom webpack config not currently supported by `obt`. Fortunately `n-makefile` supports such configuration and thus was a good time to update our Makefile to use the new shiny magic.

This simply followed the [n-makefile install steps](https://github.com/Financial-Times/n-makefile#installing) and creates a `n-makefile.json` manifest.

#### Notes
- Developers will need to `rm -rf node_modules` or `make clean` then `make install`
- The committed `n.Makefile` should never be edited and only ever updated automatically via `make update-tools`
- Had to manually add a lot of webpack/babel dev-deps that `obt` gave us for free, but should be a one-off.
- Had to temporarily disable `verify` linting task as new rules are much stricter and we have a lot to clean up :grimacing: 

Thanks to @matthew-andrews & @ironsidevsquincy for their help :)


keirog:
:+1: :sparkles:

laurieboyes:
Late to the party, but should we be storing generated files like `n.Makefile` in the repo?

matthew-andrews:
@laurieboyes normally, no, but that's currently how `n.Makefile` is architected.  the benefit is that the repo is self contained — you should be able to check it out on any Mac, run `make install` and it'll work

matthew-andrews:
got some thoughts here https://github.com/Financial-Times/n-makefile/issues/22

laurieboyes:
Ah I see cool, thank you

phamann:
Shipping! 🚢

matthew-andrews:
boom

PR #673, opened 2016-04-01
Fix empty card styling

keirog:
Before (big) | After (big)
---|---
before-big | after-big

Before (medium) | After (medium)
---|---
before-medium | after-medium


laurieboyes:
:+1:

PR #674, opened 2016-04-01
Swap n-topic list cards for n-cards

keirog:
Fairwell, n-topic! :dizzy: 

**Before:**
before

**After:**
after


laurieboyes:
:heart_eyes: :+1:

PR #677, opened 2016-04-04
Collections filtering

laurieboyes:
Don't show collections from which user already follows >= 50% of topics


server/lib/filter-collections.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+const numberRequired = 4;
+
+export default (rawCollections, follows) => {
+
+	const filteredCollections = rawCollections.filter(collection => {
+		const thisCollectionConceptsFollowed = collection.concepts.filter(concept => {
+			return follows.filter(followedTopic => concept.name === followedTopic.name).length;
```
phamann: I'd use `find` or `some` here instead of filter.

test/server/filter-collections.spec.js
```diff
@@ -0,0 +1,36 @@
...
+'use strict';
+const expect = require('chai').expect;
+
+const lib = require('../../server/lib/filter-collections');
+const rawCollections = require('../fixtures/raw-collections');
+const followsEuLessThanHalf = require('../fixtures/collection-eu-less-than-half');
+const followsEuMoreThanHalf = require('../fixtures/collection-eu-more-than-half');
+
+
+describe.only('Filter collections', () => {
```
phamann: :eyes: :grimacing:

phamann:
:+1: Other than comments.

PR #682, opened 2016-04-06
Clientside react rendering

phamann:
This refactors our react output to enable client-side instantiation/rendering of server-rendered components. I've used the topic collection section as an example.

- Bump version of `n-section`
- Update section layout config to use new `type` string format
- Update `react-render-to-sting` handlebars helper to switch format based on `isAlive` prop
- Out raw json data for component in view if `isAlive` prop is true
- Update client-side `explore/index.js` to re-init component


client/components/explore/index.js
```diff
@@ -20,4 +24,13 @@ export default function () {
 	});
 
 	destinationEl.addEventListener('myft.ui.topicSearch.updated', () => myFtUi.updateUi(destinationEl));
+
+	[].slice.call(document.querySelectorAll('.myft__recommendation')).forEach(section => {
```
laurieboyes: Could this use `Array.from` instead?

keirog:
:+1:

PR #687, opened 2016-04-07
give main element block styling across myFT page for IE10's sake

laurieboyes:
Fixes https://trello.com/c/qHsK0B4t/641-ie-11-10-topic-cards-have-been-stretched-to-display-the-full-screen

Not sure whether or not this is the right level at which to be this opinionated. Any thoughts welcome.


keirog:
Would prefer it to live here: https://github.com/Financial-Times/n-normalize/pull/3

keirog:
(myft-page is the only app using a `<main>` at the moment but hopefully won't be soon)

laurieboyes:
Ah yessss nice! I didn't know we had n-normalize

PR #689, opened 2016-04-07
don't show collections section header when collections flag off

laurieboyes:
I.e. never do this:

screen shot 2016-04-07 at 15 00 40


keirog:
:scream: :+1:

PR #690, opened 2016-04-07
stop occasional collections section duplication

laurieboyes:
Was mutating a list retained in memory between requests.

Todo
- [ ] Take a good hard look at myself


keirog:
Ha, :+1:

PR #691, opened 2016-04-08
Always show a topic collection empty state / feedback card

laurieboyes:


server/lib/deepCopy.js
```diff
@@ -0,0 +1 @@
+export default (value) => JSON.parse(JSON.stringify(value));
```
laurieboyes: Happy for a discussion on whether or not this is evil. My position is that it's not, but maybe I could do with more descriptive module naming so it's clear how the copy is implemented?

server/lib/filter-collections.js
```diff
@@ -1,11 +1,13 @@
...
+import deepCopy from './deepCopy'
 
+// i.e. 3 and an empty one
 const numberRequired = 4;
 
 function resultOrExcess(data, excess) {
 	let result;
 	if(data.length >= numberRequired) {
 		result = data.slice(0, numberRequired);
+		result[3].isEmpty = true;
```
keirog: Sneaky feature addition

server/lib/deepCopy.js
phamann: https://facebook.github.io/immutable-js/ :trollface:

server/lib/filter-collections.js
laurieboyes: Whoops! ~~Sneaky feature addition~~ Forgetful PR naming

PR #695, opened 2016-04-08
Refactor recommendation section into own partial

phamann:
- Move recommendation template and logic into own partial to be reusable across views
- Use a `js-` class as hook for React initialisation on the client side
- Move client initialisation logic into own module


laurieboyes:
:+1:

PR #702, opened 2016-04-12
Revive verify (reviverify?)

keirog:
Rhys pointed out that we weren't actually running `make verify` anymore, so this brings it back (and fixes the issues that crept in).


laurieboyes:
👍

PR #703, opened 2016-04-12
enabled smoke tests again

laurieboyes:
Make `smoke` a part of `provision`
Removed redundant following page fragment tests
Fixed following tests for layout now appearing in URL


keirog:
👍 😬

PR #715, opened 2016-04-25
Use latest n-section

ironsidevsquincy:
Which uses latest `n-card`, which uses latest `n-image`, which uses `imgset` and `sizes` for attributes for responsive images (rather than the `<picture />` element


phamann:
👍  Thank you for this. Do you think it warrants a once over form QA first?

ironsidevsquincy:
Have tested, believe it's safe to merge

PR #717, opened 2016-04-29
Add: o-card class to empty card partial

andygout:
cc @phamann @laurieboyes @keirog 

Will ensure that placeholder card text is vertically-centred (as it is currently, pictured) once we switch to `o-card`.

And as soon as build is passing...

screen shot 2016-04-29 at 10 28 42


laurieboyes:
nice 👍

PR #719, opened 2016-05-03
Bump n-layout

keirog:
Nothing to see here


ironsidevsquincy:
Will you need to turn off silent mode?

keirog:
Ack, thanks

PR #721, opened 2016-05-03
check user ID is valid UUID before accepting

laurieboyes:
Depends on this PR https://github.com/Financial-Times/next-myft-client/pull/88 because of the way I did it (came to the solution a bit backwardly)


keirog:
👍

PR #722, opened 2016-05-04
Downgrade n-image to get template back

keirog:


keirog:
Our product tour was breaking with the latest n-image, as the handlebars template we were using was removed (it's now React-only). I'll bump it back up and amend our templates after getting our builds healthy again.

PR #725, opened 2016-05-04
Bump back up to n-image v3

keirog:
- bumps back up to n-image v3, but this time…
- amends our image models/templates to use the now React-only image component

Visual differences: none


laurieboyes:
👍

PR #726, opened 2016-05-05
Temporarily skip some tests

keirog:
I know. Sorry world.


phamann:
👍 Jst to unblock the other stuff, but would be good to know what this is actually related too

keirog:
Ta, and yep definitely

PR #730, opened 2016-05-11
Allow adding properties to newly created relationships and add one for email preference default timezone

laurieboyes:
Achieves this in core by looking for properties in the body that are prefixed with `_rel.` and putting them into a nested `_rel` object, which is where the API expects to find relationship properties.

Depends on https://github.com/Financial-Times/next-myft-ui/pull/128


phamann:
👍 LGTM

laurieboyes:
I guess I'm proposing we name our timezones as follows:

EU = Europe and Africa
US = Americas
AS = Asia

server/controllers/alerts.js
```diff
@@ -23,6 +23,10 @@ module.exports = (req, res, next) => {
 	userPrefsPromise
 		.then(prefs => {
 			viewModel.hasPreferred = prefs;
+			viewModel.emailRelProperties = [{
+				name: 'timezone',
+				value: 'EU'
```
phamann: As discussed, would prefer if the value was an IANA timezone as used by [`moment-timezone`](https://github.com/moment/moment-timezone)

PR #733, opened 2016-05-12
persiting query string in qr code middleware

wheresrhys:
@Financial-Times/next-myft


phamann:
👍 You have failing tests

PR #734, opened 2016-05-17
Gracefully handle broken/null graphql queries

laurieboyes:


server/lib/sanitize-explore-topics.js
```diff
@@ -46,13 +46,23 @@ function dedupe(data) {
...
 	const deduped = dedupe(withoutFollows);
 
 	const result = keys.map((key, i) => Object.assign({}, { [key]: deduped[i] }));
 
+	//put nulls back in
+	rawData.forEach((d,i) => {
+		if(d === null) {
+			result.splice(i, 0, null);
+		}
+	});
```
phamann: Why do you put them back in?

laurieboyes: So the results match up with the inputs. Probably need a better comment there

PR #736, opened 2016-05-17
Link ‘Manage topics’ to a page that you can actually use to manage topics in the core experience

matthew-andrews:


laurieboyes:
👍  yay

PR #738, opened 2016-05-20
update to n-ui release candidate

wheresrhys:
@Financial-Times/next-myft 

Note that this also disables critical path css - It was about 20k gzipped & minified!

Will revisit once one or both of the following are done
- support generating different critical css per page
- skeleton styles implemented for styles which _coudl_ be deferred


laurieboyes:
👍

ironsidevsquincy:
We were happy for this to go into the head, as it would still give a perf boost. Is there a reason not to?

wheresrhys:
Myft page isn't being worked on much these days, so there's possibly greater benefits to be had through caching the css for repeat visits; 20k doesn't fit in the first http package.

I don't have a strong opinion though. Unless someone is very strongly in favour of keeping head.css I'll - quite literally - toss a coin

ironsidevsquincy:
I'd say keep it for consistency, and as an impetus to move stuff into main

PR #773, opened 2016-06-20
Remove: Extraneous query params from search suggestions request

andygout:
cc @laurieboyes 

[`next-tme-search-api`](https://github.com/Financial-Times/next-tme-search-api/blob/master/server/controllers/suggestions.js) now returns a one-dimensional array; therefore `flatten` query param is no longer required and so has been removed.


laurieboyes:
As per the other one 👍

PR #774, opened 2016-06-20
Refactor surrogate-key values to be HTTP compliant

phamann:
This fixes our surrogate keys to remove unsafe characters from the key value, such as `:` and a `,`. Whilst these limitations are not detailed on the Fastly `surrogate-key` docs they definitely fix our caching problems. I will follow up with Fastly directly to discuss our findings.

See [this](http://www.bizcoder.com/everything-you-need-to-know-about-http-header-syntax-but-were-afraid-to-ask#footnote3) article and header syntax table for further info.

#### Note:
This needs to be deployed alongside a similar change to the `PURGE`'ing inside [next-myft-api](https://github.com/Financial-Times/next-myft-api/pull/224).


keirog:
🏆 👍

PR #777, opened 2016-06-22
Explicitly turn critical CSS off in IE8

ironsidevsquincy:


keirog:
Does that mean it'll fall back to just loading everything in one file?

ironsidevsquincy:
It does already, due to this - https://github.com/Financial-Times/n-webpack/blob/master/lib/index.js#L99
Any file that isn't `main.css` has critical compilation turned off automatically. I wan't to remove that, to allow us to have multiple `main`-named files that are compiled, so before that can happen need to explicitly turn off critical compilation for IE8

wheresrhys:
would it be less magic to add a separate webpack build, with headCss turned off, for ie8 to each app (relatively easy with n-webpack now - can just export an array of configs)?

ironsidevsquincy:
Hmmm, not sure that would be less magic. We can already define whether a file outputs critical css or not in the `configure` stage in the sass, want to get away with doing it opaquely in two separate places

wheresrhys:
Maybe get rid of the `critical` option in the sass altogether then, and always output the comments? (though leave in the ability to configure different components as critical/non-critical). Maybe magic is the wrong term, but this would leave you needing to do 3 things in 3 places, and create a new sass variable, in order to build and deliver critical css. Moving the critical on/offness into n-webpack config reduces it to 2 with simpler sass, and as it'd all be in js there'd be potential for simplifying further (I quite like the idea of taking over n-makefile.json to define the options to pass in to n-ui, n-webpack and, quite possibly, res.locals in express too)

ironsidevsquincy:
I think the problem is we're abstracting it too much. I think the current setup makes sense, the sass file itself decides whether it needs to create a critical css; it's very clear and explicit, and gives us the flexibility to decide which files should go critical, rather than happening behind the scenes. Don't think this is really adding any complexity (just setting a new variable)

ironsidevsquincy:
Which isn't to say we can't go down that route in the future, but let's go simple for now and stick with what with got (just consolidate the two places that define whether critical-ness is applied into one), and see how far it gets us

PR #779, opened 2016-06-24
Add new myFT digest preference

keirog:
Probably easier to review each commit 😬 
Turn on the `myFtDigestPrefs` flag to see the changes in /alerts.

- Core: Allows form method override in the request body

    This lets set an off state for non-binary preferences like the new
    digest pref, where there are three submit buttons per form: daily,
    weekly, and off.
    Each submit button submits a different name/value, where the off button
    sends `name="method"`, `value="delete"`. (Daily/weekly buttons send a
    `_rel.selected` property.

- Core: Adds messaging for new pref for core journeys
- Tidier mark-up spacing and heading hierarchy, for own sanity

big


laurieboyes:
Looks solid to me, nice solution 👍

PR #787, opened 2016-06-28
Hopefully reduce graphql-api timeout errors

keirog:
The default is 4 secs; have changed to 10 secs


laurieboyes:
👍

phamann:
I'm afraid this won't have any effect, both Fastly and our router have a hard cut off of 3 secs. Therefore the connection will be closed before this can finish.

keirog:
Oh, damn ok ta

phamann:
The solution is to fix the underlying slowness of the service graph-ql is hitting, not just up the timeout 😺

ironsidevsquincy:
The graphql api isn't behind fastly, but yeah, the router timeout will kick in

keirog:
I thought it wouldn't do any harm as it's polling, but didn't realised about the router timeout. Drat

PR #789, opened 2016-06-28
Update follow-email to work with the new digest

keirog:


laurieboyes:
👍

server/controllers/follow-email/confirmation.js
```diff
@@ -10,8 +10,12 @@ module.exports = (req, res) => {
 	res.render('confirmation', {
 		layout: 'wrapper',
 		message: coreUserMessage(req.query.method, 'preferred', res.locals.userId, {
-			requestBody: {},
-			subject: 'email-daily-digest'
+			requestBody: {
+				'_rel': {
+					type: 'daily'
```
phamann: Don't we need to have a `timezone` property here too?

keirog: Not at the moment at least—it's only using this data to render the core confirmation message, which doesn't mention it.

Just a lazy reuse of the `coreUserMessage` function without refactoring it to be less opinionated about object structure 😬

PR #790, opened 2016-06-29
Remove segment ID tracking from /products link on the product tour

phamann:
As of next week B2B customers are going to be directed to the product-tour page. For anonymous users the signup link points to `/products` which is correct, but contains a B2C `?segmentId` query param for attribution tracking, this isn't ideal for B2B journeys. Therefore, until we can `Vary` on b2b/b2c users the diplomatic solution is to remove the tracking entirely.


laurieboyes:
Sure 👍

commuterjoy:
Also segment ID is irrelevant for steps that don't preceed a direct entry in to the signup form :)

PR #800, opened 2016-07-04
Allow us to turn the follow-email feature back on just for enhanced users

keirog:
There's a fastly-related bug somewhere that only affects core users,
and we want to get this feature back on as soon as possible.


laurieboyes:
👍

PR #801, opened 2016-07-04
Core experience redirect from mutation

laurieboyes:
Mutating core-experience routes redirect to a safe GET-requested confirmation page instead of rendering HTML themselves


server/controllers/core-action-confirm.js
```diff
@@ -0,0 +1,13 @@
...
+
+export default function (req, res) {
+
+	const actionKey = req.params.actionKey;
+	const userId = res.locals.userId;
+	const opts = req.query;
+
+	res.render('confirmation', {
+		layout: 'wrapper',
+		message: getConfirmationMessage(actionKey, userId, opts)
```
phamann: Can we throw a `404` if the action doesn't map to a predefined message?

client/main.scss
```diff
@@ -110,15 +110,15 @@ $output-critical: true !default;
...
-			@include oTypographySansBold(xl);
-		}
 	}
 	a {
 		@include nLinksBody();
 	}
 	b {
 		@include oTypographySansBold(l);
+		@include oGridRespondTo(M) {
+			@include oTypographySansBold(xl);
```
keirog: Yessss 💯 

You could take it a step further and use `oTypographySansBoldSize` here

phamann:
:+1: LGTM 

Would be good to put more thought to a proper Redis backed flash messaging system.

keirog:
👍

server/controllers/core-action-confirm.js
```diff
@@ -1,13 +1,21 @@
...
 	const opts = req.query;
 
-	res.render('confirmation', {
-		layout: 'wrapper',
-		message: getConfirmationMessage(actionKey, userId, opts)
-	});
+	try {
+		res.render('confirmation', {
+			layout: 'wrapper',
+			message: getConfirmationMessage(actionKey, userId, opts)
```
keirog: Seems like calling and checking `getConfirmationMessage` first would be better, before calling render, no?

PR #806, opened 2016-07-07
Fix edit-topics button styling

keirog:
I think something changed in n-ui that meant we’ve stopped outputting the o-expander CSS, but since it only requires 2 lines of CSS to fix, let’s favour that.

Before:
screen shot 2016-07-07 at 15 01 12

After:
screen shot 2016-07-07 at 15 01 46


laurieboyes:
👍

PR #807, opened 2016-07-07
Add middleware logging for bug hunting

keirog:
Trying to find out why the follow-email post route isn't receiving a user UUID (but _does_ have a post body)


laurieboyes:
👍  looks bon to me

PR #808, opened 2016-07-07
Some product tour IE fixes

laurieboyes:
Some before shots:
screen shot 2016-07-08 at 11 28 12

Some after shots:
screen shot 2016-07-08 at 11 29 21


keirog:
👍

PR #810, opened 2016-07-08
Bower resolutions

keirog:


laurieboyes:
😬 👍

leggsimon:
Shouldn't the new n-ui resolve these for you 😨 ??

keirog:
Ah thanks, was committed on the same minute! Much better, thanks @leggsimon

PR #818, opened 2016-07-11
“Fix Middlewoes”, or “OMG, It Was You the Entire Time”

keirog:
So… it turns out we’ve been mounting all routes to both the apiRoute *and* authRoute routers. 😱 

This PR corrects the problem and attempts to make the app.js file more reflective of what happens at runtime.

I highly recommend viewing this PR one commit at a time 😬 

cat makes a discovery


server/middleware/auth/public.js
```diff
@@ -2,13 +2,18 @@
...
 const logger = require('@financial-times/n-logger').default;
 const whitelist = [
-	/^\/myft\/product-tour(?!-get-started)/
+	/^\/myft\/product-tour(?!-get-started)/,
+	/^\/(__)?myft\/list\/([^/]+)$/
 ];
 
 module.exports = (req, res, next) => {
-	logger.info('In public middleware', { event: 'BUG_ADVENTURE', url: req.url, originalUrl: req.originalUrl });
+	logger.info('In public middleware', { event: 'BUG_ADVENTURE', method: req.method, url: req.url, originalUrl: req.originalUrl });
```
phamann: Do we still need this debug log?

phamann:
Woohoo! But the suspense is killing me, did it fix your core journey bug?

phamann:
👍 💯

server/middleware/auth/public.js
keirog: Just leaving this in until everything is confirmed fixed in prod

keirog:
I‘m *really* hoping so… I can‘t be sure until merging

keirog:
(For those following along, it didn‘t fix it but did make it _much_ easier to find out was going on)

PR #819, opened 2016-07-12
Middleware wiggle

keirog:
Preflight-skipped requests trigger the `res.locals.anon.userIsAnonymous`
flag in n-express, which we use in the anon middleware. Therefore we
need to check for a session token *before* checking for anon state (if
we even need to check for anon still, but that's another story)


laurieboyes:
👍

PR #820, opened 2016-07-12
Speed-up layout gets

keirog:
There's a common error in Sentry for myft-page about an undefined view decorator.
I luckily caught it in the act locally yesterday, and it seemed to be because Map.get took too long to return the layout string. I really don't understand how that was possible, but let's see if
changing this makes the error go away.


keirog:
Got an IRL 👍  from @laurieboyes

PR #824, opened 2016-07-13
New alerts page layout is driven by section and layout config etc

laurieboyes:
And it currently looks like this:

screen shot 2016-07-13 at 15 54 22


keirog:
Apart from the build error (needs an alerts/main.scss), 🚢  🚢  🚢

PR #825, opened 2016-07-14
fix my-newsletters layout requiring variable number of components

laurieboyes:
Proposed solution is to give the option of making layout definitions functions that can refer to an `opts` object


keirog:
Seems like a good workaround without refactoring the whole thing 👍

PR #826, opened 2016-07-14
Product tour optimisations with will-change

laurieboyes:
This extra property prevents the product tour animations requiring repaints [in modern browsers](http://caniuse.com/#search=will-change).

Verified with the rendering profiler (saw a good talk):
screen shot 2016-07-14 at 12 34 14

Also removed some no-longer-used stuff.


keirog:
👍  💯

PR #828, opened 2016-07-15
Point to newsletter images hosted in next-assets (decoupling from email platform)

laurieboyes:
Also added brexit briefing newsletter details


keirog:
👍

PR #831, opened 2016-07-18
Track the myFT link in our header

keirog:


debugwand:
looks ok, is there not a myft tracking identifier higher up to define the whole page as myft (might not be but I am assuming there is something to demark myft pages)? this is specifically clicking on the logo to go to /myft - maybe ask James if he's the one using beacon to find this bit of info

keirog:
Yep about the higher-up identifier, it’s on the body as `page:myft-page/[following|saved-articles|explore|alerts]`.

Our general approach has been to add identifiers to every link in Next, but I forgot to add it for this one 😳  . Adam B made [a Chrome extension to tell us off when we forget](https://chrome.google.com/webstore/detail/mollydobbin-tracking-tool/cmhfmieofijgphlefokpckljhgcmaneb) but I had it turned off 😬

PR #836, opened 2016-07-18
Replace deprecated n-buttons usage*

keirog:
*Sass wise, all n-buttons ever did was modify o-buttons’ standard text size. Now we’re just doing that by [modifying the `$o-buttons-sizes` Sass map](https://github.com/Financial-Times/n-ui/blob/master/buttons/_mixins.scss#L2), and n-buttons is deprecated. Hopefully, one day we’ll be able to stop overriding this at all 😬


laurieboyes:
👍

PR #839, opened 2016-07-18
update list styles for n-card changes

laurieboyes:
Some of the styles that were previously specific for concept card have been moved to the more generic myft card


keirog:
👍

PR #841, opened 2016-07-20
More newsletter styles

laurieboyes:
Getting closer:
screen shot 2016-07-20 at 14 26 50


shared/components/alerts/newsletter/_newsletter.scss
```diff
@@ -1,9 +1,30 @@
...
+	margin: 0;
 
 	dd {
 		margin-left: 0;
 	}
 }
+
+.card__newsletter-header-meta {
+	text-transform: uppercase;
+	color: oColorsGetPaletteColor('blue');
```
laurieboyes: I didn't think this was worth a use case

shared/components/alerts/newsletter/newsletter.js
```diff
@@ -24,34 +24,40 @@ export default class extends Component {
...
+					<dl className="card__newsletter-frequency card__myft-newsletter__data">
+						<dt className="n-util-visually-hidden">Frequency</dt>
+						<dd>{item.frequency}</dd>
+					</dl>
+					<form className="n-myft-ui card__newsletter-unsubscribe" action={unsubscribeAction} method="POST">
+						<input type="hidden" name="listName" value={item.name}/>
+						<button
+							className="n-myft-ui__button"
+							aria-label={`Unsubscribe from ${item.name}`}
+							title={`Unsubscribe from ${item.name}`}
```
debugwand: https://www.paciellogroup.com/blog/2012/01/html5-accessibility-chops-title-attribute-use-and-abuse/ you would be better off putting hidden text after 'Unsubscribe' in the contents of the button

keirog: Not sure I agree here, at least about the `title` being replaced. I like to think of modern `title` usage as a heartfelt apology for us focussing too much on visual minimalism </🎭>. I agree we are abusing it, but only in lieu of winning a long battle with designers.

The button (IMO totally incorrectly) only visually states the verb but not the object. The current solution—although not great—caters for:
- cursor-input users via `title`, although only as a hope that they’ll roll over if confused
- screen-reader users via `aria-label` (I’m assuming, maybe incorrectly?)

The big flaw is there’s no added context for touch-input folks, but that’s only fixed visually (AFAIK).

If we remove the `title` and just use hidden text, we could then remove the `aria-label` but would lose the (hopefully helpful) context for cursor-input users.

shared/components/alerts/newsletter/_newsletter.scss
```diff
@@ -1,9 +1,37 @@
-.card__newsletter-header-meta {
-	text-transform: uppercase;
-	color: oColorsGetPaletteColor('blue');
-	margin: 0 0 0 10px;
+@include oColorsSetUseCase(myft-newsletter, all, 'teal-1');
```
keirog: Maybe not worth a use-case here?

keirog:
👍

shared/components/alerts/newsletter/newsletter.js
debugwand: the table at the bottom of this page http://www.deque.com/blog/text-links-practices-screen-readers/ (albeit done with links not buttons) shows what different tools do, as you can see its a lot like when we had to deal with wildly differing browsers

some users get both title and text, some get both label and text, some get both label and title, some get just text, some get just label

laurieboyes: Not trying to shut down debate but going to merge for now as this is a wider issue anyway (affecting all our buttons)

debugwand: appreciate that it is helpful for sighted users' usability though.  sometimes assistive tech limits do leave us in a position where we have to 'pick' which is the least bad experience because different user groups have directly opposing needs

keirog: Oof, that’s a sad read. Is there a solution that could cater for everyone, that anyone can think of? (Short of changing the visual design 😞 )

keirog: (Unless I’m just being overly sensitive about the buttons lacking context, in which case ignore)

debugwand: a tooltip rendered by us rather than the browser. that might be possible with css-only. is there an origami component?

keirog: for touch users too though?

debugwand: Touch users wouldn't get title either, so they are in the same situation as before. Visually, the container box and title at the top should give visual users context of what they are unsubscribing from - if not the design needs tweaked.

Changing the button text - more of a product / design / content / house style decision. I have a personal preference for more explicit labels that make sense out of context - that's an opinion. We can provide support for the other more implicit approach.

Screenreader users need something a bit more explicit because a) its slower to listen all that stuff and maintain context in your head, and b) the way they tend to skip over content and jump between interactive elements  rather than listen to the whole page (same reason that 'view more' is not good link text)

keirog: I guess overall we should cater for the minority in this case, as fully-sighted folks will (unfortunately for others) be the only people we’re likely to do user testing on. Data from that testing should hopefully show if the visual designs don’t give enough context for the majority. Ok I’m on board with your suggestion 👍

keirog: Sorry for turning this PR into a forum!

PR #849, opened 2016-07-22
Use auth token not uname/pwd in url

debugwand:
https://github.com/Financial-Times/next-config-vars/pull/293

and remember to update your local .env


phamann:
👍

PR #850, opened 2016-07-22
revert following page -> saved articles page redirect

laurieboyes:
Essentially reverts this: https://github.com/Financial-Times/next-myft-page/pull/838

This was a bad idea because it means if you get redirected to the saved articles page when you don't following anything, the following page link is basically just a refresh button 😬 

Also, Patrick pointed out that redirects are now cached, so having an extra redirect for the intelligent myFT routing isn't such a bad thing after all.

EDIT: This PR now also removes the redirect from the following page so that if you follow no topics, you _can_ access the empty-state following page if you click the following link from the myFT subnav (previously was an obnoxious redirect to product tour). All lack-of-engagement-based redirects are now handled at /myft root level


debugwand:
👍

laurieboyes:
This PR now also removes the redirect from the following page so that if you follow no topics, you can access the empty-state following page if you click the following link from the myFT subnav (previously was an obnoxious redirect to product tour)

PR #855, opened 2016-07-26
Prevent upgrade to 2.15.0

debugwand:
Just in case removal of 'delete' from the handling of buttons in n-ui ui.js causes us problems, prevent upgrade to n-ui 2.15.0


keirog:
Not sure about this—tildes have been the source of a lot of pain. Is this just a precaution in order to ship this evening?

debugwand:
It is a precaution, will go back to caret when moving to 2.15.0

PR #859, opened 2016-07-27
Don’t cache the new alerts page

keirog:
Newsletter subscription changes are external to *my*FT land, so we can't
cache this page. Alternative solutions quite welcome!


laurieboyes:
👍  I'm on board with this solution, definitely for the short term anyway

PR #862, opened 2016-07-27
Fix on/off prefs for core

keirog:


keirog:
Merging this in as it's broken in prod

PR #863, opened 2016-07-28
Add missing functionality to react on/off prefs

keirog:
This is to match some of the existing functionality of the non-react on/off prefs


laurieboyes:
👍

PR #873, opened 2016-08-01
Cleanup BEM classname conventions on newsletter cards

phamann:
Tidies up some of the BEM classnames on the newsletter cards. Essentially `.card__myft-newsletter` => `.card--myft-newsletter` Still not happy with the `myft-` prefix on most child elements, but that's all contained with n-card so am going to leave for the time being.


laurieboyes:
👍

PR #881, opened 2016-08-02
Remove old handlebars dependency

keirog:
We don’t appear to be using this, and it’s causing VersionEye to flag
myft-page up as a vulnerable app because of a hole in handlebars v3
(n-express uses v4).


laurieboyes:
👍

PR #886, opened 2016-08-03
Works around non-flexbox layout dodginess 😁

keirog:
Obviously not ideal, but gets the job done. Might need a tweak as the
digest card matures.


client/main-ie8.scss
```diff
@@ -17,3 +17,7 @@ $output-critical: false;
 	}
 
 }
+
+#my-newsletters-section-content .card--myft-newsletter {
```
laurieboyes: Why ID and not class?

keirog: Oops, thanks, shouldn't be there

client/main-ie8.scss
```diff
@@ -17,3 +17,7 @@ $output-critical: false;
 	}
 
 }
+
+.card--myft-newsletter {
+	height: 370px;
```
laurieboyes: Also, sorry, should have spotted this a minute ago. Is it worth making this a min-height in case the card really does need to be bigger (a very long description for example)? Very unlikely edge case to feel free to ignore

keirog: height = min-height in IE 8

keirog: Oh hang on… was that IE 6

keirog: Apologies, it was 6. Got mixed up

laurieboyes:
👍

PR #887, opened 2016-08-03
More digest styles

laurieboyes:
Getting there slowly

Before:
screen shot 2016-08-03 at 10 19 20

After:
screen shot 2016-08-03 at 10 19 24

(depends on https://github.com/Financial-Times/n-ui/pull/322)


shared/components/alerts/myft-digest/main.scss
```diff
@@ -10,3 +10,9 @@
 	}
 
 }
+
+.card__myft-frequency__change {
```
keirog: How do you feel about singularising the BEM elements down to just `myft-frequency-change`?

keirog:
👍

shared/components/alerts/myft-digest/main.scss
laurieboyes: I guess I'd feel good about it if it only containing component styles and no margin/positional styles. Maybe I could split it out into `myft-frequency-change` for the button stuff and `card__myft-frequency__change` for the margin??

PR #888, opened 2016-08-03
Even more digest styles

laurieboyes:
Getting closer. Depends on https://github.com/Financial-Times/n-ui/pull/324.

Before (core):
screen shot 2016-08-03 at 12 02 30

After (core):
screen shot 2016-08-03 at 12 02 26

Note that on enhanced, the change button now juuust about fits:
screen shot 2016-08-03 at 12 04 08


keirog:
👍 👍 👍

PR #889, opened 2016-08-03
Made myft-digest prefs setting more exciting

laurieboyes:
new

Core remains intact


client/main.scss
```diff
@@ -21,7 +21,7 @@ $output-critical: true !default;
 @import "n-section/main";
 
 @include oColorsSetUseCase(myft-page-cold, all, 'cold-1');
-@include oColorsSetUseCase(myft-newsletter, all, 'teal-1');
+@include oColorsSetColor('myft-newsletter-preference-background', #F4EADA);
```
laurieboyes: Pretty sure this is criminal but put it in for the sake of discussion. In the design, this colour was an origami colour with an opacity. This was fine until I realised we had to overlay this other panel with the same colour - doh!

It definitely doesn't seem worth having this colour in origami, or even in n-ui, as it'll only be used for myFT. Does anyone have any thoughts?

keirog: Although not perfect, you could do something like this: https://github.com/Financial-Times/n-card/pull/47/files

wheresrhys: Can't we change the design? Saying yes to using new colours for a single use case is always a slippery slope

keirog: It’s technically not a new colour though, if I‘m understanding correctly… It’s more to do with opacity issues

laurieboyes: Yep, we're back on teal-1

keirog:
Might be worth checking with mattH/designux channel (if you haven’t already) that the animation fits with site-wide conventions. Same for the close button I guess.

Otherwise 👍 ✨

PR #893, opened 2016-08-04
hopefully fix weird character after freq

laurieboyes:
Explicit space doesn't appear to be necessary and seems to be causing this on clientside prod:
screen shot 2016-08-04 at 15 22 11


keirog:
👍

PR #895, opened 2016-08-04
updated list of automated newsletters

laurieboyes:
Did some `querySelector`ing on the nbe page to get a more up-to-date list


keirog:
👍

PR #898, opened 2016-08-05
fix subscribe updating ellipsis getting double escaped

laurieboyes:
Went for the 'last resort', as giving the unicode char didn't work for me: https://facebook.github.io/react/docs/jsx-gotchas.html#html-entities


keirog:
👍

shared/components/alerts/newsletter/index.js
```diff
@@ -32,7 +32,7 @@ export default class extends Component {
 	}
 
 	getSubscriptionText(state) {
-		if(state.processing) { return 'Updating…'; }
+		if(state.processing) { return 'Updating&#8230'; }
```
debugwand: give \u2026 a try - that is the JS-friendly way of escaping UTF-8 characters

laurieboyes: Ah yeah I didn't have any luck with that before but I've just tried again and found it's cause babel's stomping on my `\u`s! There must be a way around that. Watch this space.

laurieboyes: Turns out it's not babel causing the problem. Thought it might be n-webpack but haven't managed to find conclusive proof in my timeboxing window so raising an issue and committing this for now.

PR #899, opened 2016-08-05
Change IE8 layout to a more brittle but visually better solution

keirog:
This clears the floats for every row (IE 8 is fixed to the XL layout), in an obviously brittle way but it gets the job done


laurieboyes:
I love it 👍 ❤️

PR #904, opened 2016-08-08
Remove dependency on old flag

keirog:
We’re not using the `myFTPushNotifications` flag anymore, so will delete it


laurieboyes:
👍

PR #905, opened 2016-08-08
Use local session decoding component

wheresrhys:
Should improve response times and reliability for POST and DELETEs @Financial-Times/next-myft


keirog:
😻  👍

PR #912, opened 2016-08-09
Move chai to deps from devDeps

keirog:
There seems to be a third-party package that is naughtily(?) requiring
Chai as a dependency and not devDependency, so when we prune, it breaks
our build-production step


laurieboyes:
👍

PR #922, opened 2016-08-18
Collections

debugwand:


bower.json
```diff
@@ -2,7 +2,7 @@
   "name": "next-myft-page",
   "dependencies": {
     "n-myft-suggested-topics": "^3.0.0",
-    "n-section": "^5.2.2",
+    "n-section": "^6.1.3",
```
laurieboyes: Do we know what the reason for the major bump was? I've looked at the release notes for v6.0.0 and it doesn't give many clues

client/components/recommendations/index.js
```diff
@@ -1,13 +1,13 @@
 import React from 'react';
-import { render } from 'react-dom';
+import { render } from 'preact';
 import { Section } from '@financial-times/n-section';
-
 export default function () {
 	[].slice.call(document.querySelectorAll('.js-is-alive')).forEach(section => {
 		const data = document.getElementById(`js-initial-data-${section.id}`);
+		const replaces = section.querySelectorAll('section');
```
laurieboyes: Can't see where this is used?

debugwand: ah - it isn't. not sure why my linter isn't complaining

bower.json
debugwand: I ran this:
https://github.com/Financial-Times/n-section/compare/v5.2.2...master
and it looks like there was a lot of changes to article. I've done manual testing on our pages to check that things still work

 1195018
Edit
Article model data transformation migrated to n-content-model
@andygnewman andygnewman released this 15 days ago · 10 commits to master since this release

Most of the logic that manipulates the data to make it fit for use with n-card, has been migrated into n-content-model v2.0.0

This makes it simpler to bypass n-section and use n-card directly as well as the ability to use content data sourced from either next-graphql-api or directly from next's elastic search cluster.

laurieboyes: Ah sweet so the breaking change will have been for articles not concepts, collections etc. Good enough for me! 👍

laurieboyes:
Hooray they're back 👍

PR #926, opened 2016-08-22
Add sponsorship

keirog:
**This is fine**

after


keirog:
I've added this behind a `myFtSponsorship` flag

keirog:
BTW it only shows for L and XL breakpoints

laurieboyes:
👍

PR #927, opened 2016-08-23
Bump to node v6.4.0

keirog:
No biggie according to the [changelog](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V6.md#6.4.0); just keeping up


laurieboyes:
👍

PR #932, opened 2016-08-26
fix wonky product tour 'others' after n-image change

laurieboyes:
Before
screen shot 2016-08-26 at 13 25 30

After
screen shot 2016-08-26 at 13 25 13


keirog:
👍

PR #934, opened 2016-08-30
Instant alerts core

debugwand:


server/controllers/following/default.js
```diff
@@ -6,15 +6,15 @@ const catchTimeout = require('../../lib/catch-timeout');
...
 	//Get users followed topics
 	//Get layout preference
 	Promise.all([
 		fetchFollowed(userId),
 		getLayoutPreference(userId)
 	])
 	//Redirect to layout
-	.then(([, layout]) => {
+	.then((result) => {
+		const layout = result[1];
```
keirog: I know it's not part of this PR, but wondering why we're fetching follows and not using them

keirog:
👍

server/controllers/following/default.js
debugwand: not sure either, could it have been something that should have been deleted? or preparation for work to come? it isn't possible as far as I can see that it is getting passed on elsewhere

PR #935, opened 2016-08-31
open up the sponsorship link in a new window

VladDubrovskis:
Just making sure we do not take users away from our site.


keirog:
👎  https://mathiasbynens.github.io/rel-noopener/, https://css-tricks.com/use-target_blank/

VladDubrovskis:
fair enough :)

PR #936, opened 2016-09-01
Show info about Instant Alerts in the product tour

keirog:
With the `myFTInstantAlertsButtons` flag on:
after

Without:
before

(The myFT-page text is only linked if you have a session)


phamann:
👍

PR #938, opened 2016-09-02
Add instant alerts card on prefs page

phamann:
As discussed, this adds an instant alerts list card on the preferences page to show which alerts you are subscribed to and a call-to-action to go and mange all on the following page. 

Slightly dirty thing is, it assumes the CSS for the `autotmated-emails` list card is within scope. Thoughts?

google chrome


debugwand:
looks good

PR #939, opened 2016-09-05
Handle core-experience messaging for Instant Alert actions

keirog:


server/lib/get-confirmation-message.js
```diff
@@ -34,7 +34,11 @@ export default function (actionKey, userId, opts) {
 		case 'delete-saved' :
 			return `Article removed from your ${savedArticlesLink}`;
 		case 'put-followed' :
-			return `You followed <b>${opts.conceptName}</b>. ${manageTopicsLink}`;
+			if (typeof opts.instant !== 'undefined') {
```
debugwand: does this need feature flagged?

debugwand:
looks good

server/lib/get-confirmation-message.js
keirog: I guess we probably don't want to have flags at this point, as the user-facing bit (the button) is already flagged? It seems right to favour having our messaging match whatever’s just happened, otherwise there could be a scenario where a flag’s status changes between a user clicking “Instant alert on/off” and us responding with one of these messages.

PR #944, opened 2016-09-07
Fix iOS input box zoom on focus

keirog:
Rationale:
- on focus, iOS will zoom input boxes that have their font-size set to
< 16px (rightly so)
- `oTypographySansData(m)` is 14px
- we’re only using that mixin to grab the Metric font
- `$o-typography-sans` achieves this without clobbering over the oForms
font-size

This seems like the least bad solution.


keirog:
Merging; post-merge comments still welcome

PR #946, opened 2016-09-12
Add: Premium attribute to Topic Card articles

andygout:
cc @andygnewman @bjfletcher @phamann @laurieboyes 

First pass to quickly get Premium label up and will then apply `n-content-model` to do article transform.

- [x] *Followed Topics* cards on **Following**
- [x] *Recommended Topics* cards on **Explore** page
- [x] *Popular Topics* cards on **Explore** page (while not displaying any labels in the screengrab, they do now produce the premium attribute required to display them)
- [x] Fix tests.

#### Following
screen shot 2016-09-12 at 09 43 21

#### Explore
screen shot 2016-09-12 at 13 41 35


server/models/topic-cards.js
```diff
@@ -80,7 +80,8 @@ class TopicCards {
 				rawSrc: (article.content.mainImage) ? article.content.mainImage.url : null
 			},
 			isPodcast: article.content.metadata.some(propertyEquals('idV1', podcastIdV1)),
-			url: article.url
+			url: article.url,
+			premium: article.content.webUrl && /\/cms\/s\/3\//.test(article.content.webUrl)
```
laurieboyes: Cool. What does the `/cms/s/3/` bit represent?

phamann: @laurieboyes It's the paths that the cms generates and are used as the `webUrl` property. Sadly, this is the only way of reliably ascertaining classification. Eventually this will be exposed in the ES metadata, but in the interim we have to RegEx.

phamann:
@andygout Is it possible to prevent the label wrapping on its own and causing a orphan (as seen in Lex on your screenshot). I believe there is some CSS magic to do this.

andygout:
Unsure of what desired effect is. In the first Lex article would it be to bring 'curves' down onto second line so Premium label does not sit by itself? Wouldn't that ultimately create additional empty space on the right side of the card?

server/models/topic-cards.js
debugwand: `isPremium` would be consistent with our naming for booleans - is that possible here or are there generic dependencies?

andygout: It needs to be `premium` to be read by the concept card in `n-card`: https://github.com/Financial-Times/n-card/blob/master/components/concept.js#L74.

phamann:
> Wouldn't that ultimately create additional empty space on the right side of the card?

Yes, but i'm pretty sure if you ask a designer that would always be preferred over an orphan.

server/models/topic-cards.js
leggsimon: I think it would be good to update n-card to use `isPremium` to maintain that consistency IMHO

andygout:
Gotcha - will need to make those changes in `n-card`.

server/controllers/explore.js
```diff
@@ -38,11 +38,19 @@ module.exports = (req, res, next) => {
 		]) => {
 			const data = { fromYourIndustry, recommendedTopics, todaysTopics, popularTopics, collections };
 			const sections = getPage('explorePage', sectionContent(data, flags), flags);
+
+			const newSections = sections.map(section => {
+				return Object.assign({}, section, { data: { content: section.data && section.data.content && section.data.content.map(contentItem => {
+					const items = contentItem.items && contentItem.items.map(item => Object.assign({}, item, { premium: item.webUrl && /\/cms\/s\/3\//.test(item.webUrl) }));
+					return Object.assign({}, contentItem, { items });
+				})}});
+			});
```
ironsidevsquincy: bit hard to read this. maybe store values in variables to make it easier

server/models/topic-cards.js
andygout: @andygnewman @bjfletcher What are your opinions on this? ^^^

bjfletcher: I'm not sure why `premium` was chosen in the first place, I think probably for its flexibility as we're just developing the ideas and A/B testing... with `premium`, we have the flexibility to easily add new properties to it, with `isPremium`, we don't.
If we were to finalise the design, then, of course `isPremium` would make the most sense.
I don't mind either way. FWIW I used `isPremium` in the digest work.

bjfletcher:
@phamann do you mean, you want:

```
Andy is cool
James
Bond [Premium]
```

instead of:

```
Andy is cool
James Bond
[Premium]
```

I'm not too sure about that. When we combine last word with "Premium", in effect, creating a rather long word, causing a noticeable gap in the preceding line. 

What do you think? The "Live" classifier is orphan, by the way.

phamann:
@bjfletcher Correct the former is preferred. I only say this as I've worked in newspapers (and with print designers) for too long and have had this burnt into my brain, that we should never allow orphans. It was only a suggestion not a blocker. 

We should probably just ask one of our designers their opinion 😃

andygout:
I did speak to Oli Powell earlier and he was all for getting rid of orphans.

andygout:
Will merge this for now given it's behind a flag and any work relating to removing orphans would need to be done separately on `n-card`.

PR #950, opened 2016-09-15
Remove redundant keys

keirog:
These started causing issues with running the app locally, since updated to the latest n-makefile. Not sure why but we don't need them now as they're in config-vars.


debugwand:
looks good

bjfletcher:
👌

PR #952, opened 2016-09-15
Add: isPremium attribute to articles in Saved Articles lists

andygout:
cc @keirog @laurieboyes 

screen shot 2016-09-15 at 22 51 49


views/partials/dashboard/list.html
```diff
@@ -15,7 +15,12 @@ <h4 class="n-util-visually-hidden">Articles in this list</h4>
 	<ul class="card__concept-articles">
 	{{#each items}}
 		<li class="card__concept-article">
-			<a class="card__concept-article-link" data-trackable="article" href="/content/{{id}}">{{title}}</a>
+			<a class="card__concept-article-link card__classifier-gap" data-trackable="article" href="/content/{{id}}">
```
andygout: I wonder if it is bad form to use the [`card__concept-article-link`](https://github.com/Financial-Times/n-card/blob/9ddda771ce6c6d17bbc936b11e6bdf17b51b77bd/components/_concept.scss#L41) class from `n-card`. Similarly I have added the [`card-classifier`](https://github.com/Financial-Times/n-card/blob/602d0bcc102edc314f0581208b7e0bd6ec2770c4/components/_classifier.scss) but is it perhaps bad form to piggyback on styles that are coincidentally in the app because the component as a whole is being used elsewhere (and may be extracted at some point)?

bjfletcher: Regarding @andygout's comment above, these saved cards were already using `n-card` styling and so I think this is fine. :+1:

keirog: Agreed 👍  (The list feature is admittedly a bit of a Frankenstein’s monster)

keirog:
👍

PR #973, opened 2016-10-12
Ads sponsorship changes

andrewgeorgiou1981:
@adgad @gvonkoss @VladDubrovskis


client/main.scss
```diff
@@ -187,9 +186,13 @@ $ad-max-responsive-width: 1440px;
 		min-height: calc(100vw * 0.1734);
 		padding: 0;
 	}
+	/* for the ntt sponsorship we are temporarily using display:none for the specific pixel width range below. this can be refactored using oads config. */
```
keirog: What’s the rationale for going against the established o-grid breakpoints here?
When will the refactor into oads happen?

keirog:
Not really ok with this being merged in

PR #986, opened 2016-10-24
Email health checks

phamann:
Adds two health checks for the myft-email platform, both of which are experimental and based off cloudwatch metrics using the new check types.

Once I've seen these working for a couple of days I will add more and tweak the thresholds.


debugwand:
done elsewhere

PR #994, opened 2016-10-27
Putting next-card to bed: Part 1

keirog:
This PR has the main gubbins for moving us over from the very old next-card, to n-teasers and n-teaser-collections. To start with, we’re just replacing the time view of the myFT following page (behind the `nTeaserMyFtPage`). Other pages to come: saved-articles and the individual list page.

When we’ve migrated the 3 next-card pages over to n-teaser-collections, we can put next-card to bed. 🎉  The next step after that will be for these pages to adopt the stream-page style of n-teaser-collection, which has yet to be built.

Notable user-facing differences between next-card and the version in this PR:
- premium labels now missing (Max has given me his blessing as it’ll soon be remedied)
- standfirsts now missing (this was due to a technical complication but we can bring it back when we move to a new collection design)

**Small screen comparison:**
small

**Big screen comparison:**
big

Best viewed in separate commits 😬


client/main.scss
```diff
@@ -38,6 +39,8 @@ $ad-max-responsive-width: 1440px;
 @import "components/card-fader/main";
 
 @import "../shared/components/main";
+// Override o-teasers to look like next-card (ish), to aid migration
+@import "teaser-migration/main";
```
phamann: Why override and cause more work further down the line?

package.json
```diff
@@ -61,7 +61,7 @@
   },
   "devDependencies": {
     "@financial-times/n-heroku-tools": "^6.8.1",
-    "@financial-times/n-webpack": "^1.5.0",
+    "@financial-times/n-webpack": "financial-times/n-webpack#raw-handlebars",
```
phamann: Why pointing to a branch?

server/models/to-teaser.js
```diff
@@ -0,0 +1,7 @@
+module.exports = (c, {actions}) => ({
+	title: c.content && c.content.title,
+	url: c.content && c.content.url,
+	tag: c.primaryTag,
+	lastPublished: c.content && c.content.publishedDate,
+	actions
+});
```
phamann: I thought one of the benefits on moving to these new cards is not to have application specific modelling?

phamann:
> standfirsts now missing (this was due to a technical complication but we can bring it back when we move to a new collection design)

Interested in knowing what the technical complication was.

server/models/to-teaser.js
laurieboyes: We'll be able to chop out this modelling once I've done the work to move from personalised feed to Next API

client/main.scss
keirog: I wondered the same, but overall it seemed better (`¯\_(ツ)_/¯`) to keep the removal of next-card as a separately shippable task. The alternative would mean that its removal had a dependency on:
- the next-api rollout
- myFT template changes by the templating team
- the design/build of the site-wide stream collection

package.json
keirog: Not ideal, but this is to bring in Ben’s work here: https://github.com/Financial-Times/n-webpack/pull/36. I’ll chase for that to be merged.

server/models/to-teaser.js
keirog: This is to get something going until the next-api has been rolled out. When that’s done, we can get rid of it.

keirog:
It’s due to n-teasers no longer grouping the headline and standfirst into a CSS-hookable element for floating. This is because there won’t be any teaser collections that have a horizontal layout like next-card did. (All stream-style collections will look similar to how stream-page pages look now.)

Not sure if I’m making sense here, so maybe this helps:
next-card’s markup allowed for this group of left floats:
```
tags | headline   | date | actions
     | standfirst |
```
But n-teasers only allow this:
```
tags | headline | standfirst | date | actions
```

server/models/to-teaser.js
keirog: (^ missed Laurie’s comment)

PR #1000, opened 2016-10-28
Goodbye product-tour code

keirog:
- Splits out the clippings code from product-tour code
- Slightly simplifies clippings view rendering
- Removes remaining product-tour code

🌹


server/middleware/maintenance.js
```diff
@@ -22,13 +22,7 @@ const maintenance = (req, res, feature, options) => {
 
 module.exports = (req, res, next) => {
 
-	if ((!res.locals.flags.optInOut || !res.locals.flags.myFtApi) && /^\/myft\/product-tour/.test(req.path) && /(clippings|article-clip|shared-list)/.test(req.query.slices)) {
```
laurieboyes: Are we sure we want to lose the shared-list stuff from this?

keirog: We’ve agreed offline to get rid of this

PR #1001, opened 2016-10-31
Putting next-card to bed: Part 2 of 2

keirog:
Moves the remaining pages (saved-articles time view & individual list pages) over to n-teaser-collections.

When this is rolled out and the tyres have been kicked a bit, we can finally put next-card to bed 😄 

## Saved articles

**Small screen before/after:**

saved-small

**Big screen before/after:**

saved-big

## Individual list pages

**Small screen before/after:**
list-small

**Big screen before/after:**
list-big


client/n-ui-config.js
```diff
@@ -3,6 +3,12 @@ module.exports = {
 	features: {
 		ads: {
 			noTargeting: true
-		}
+		},
+		// Rationale for turning date off:
+		// - The new site-wide rule is to hide individual dates if the article was published more than 4 hours ago
+		// - The new stream teaser collection (not yet built) will show dates in groups (as well as individual dates for fresh/<4hrs content)
+		// - Until we have the grouping, we need to stop individual dates from hiding
+		//TODO: Remove `date:false` when new stream teaser collection is rolled out
```
laurieboyes: nice explanation 👍

server/models/to-teaser.js
```diff
@@ -1,7 +1,7 @@
 module.exports = (c, {actions}) => ({
-	title: c.content && c.content.title,
-	url: c.content && c.content.url,
+	title: c.content && c.content.title || c.headline,
+	url: c.content && c.content.url || c.id && `/content/${c.id}`,
```
laurieboyes: TIL `&&` has higher precedence than `||`

keirog: Ah, better not to have to think about it at all though—I’ll shove some brackets in for clarity

PR #1009, opened 2016-11-09
Time view uses next api

laurieboyes:
Change content source of myFT feed time view from personalised feed to shiny new Next API

Behind flag `myFtPageNextApi`

Coming soon:
- Topic view to get the same treatment
- Removal of all the code that this stuff renders redundant (squee)

## What?

Below is a bit more info on the refactoring we're trying to achieve along with the personalised feed / Next API swap out.

Next API allows us to return the content in the format of our choosing, rather than a fixed schema, which means we can sack off a lot of code and complexity that was previously used for data massaging.

### myFT following page rendering: The now (to some degree of detail, as far as I understand it)
myft page rendering

### myFT following page rendering: The future
myft page rendering v2

### A note on our graphQL queries
When n-teaser is a bit more mature, it'll have fragments for the content in the format it needs to be displayed with n-teaser. These aren't ready yet. The `nTeaserContent` fragment included here returns the content in the format n-teaser needs in its current form (but this will change).


server/controllers/following/layout.js
```diff
@@ -1,56 +1,60 @@
...
+		title: 'myFT &mdash; Following: FT news on topics you follow',
+		trackablePageName: 'following',
+		isFollowingPage: true,
+		isOwner: true,
+		noItemsMessage: 'No new articles from your topics.',
+		referrerTracking: '#myft:my-news',
+		flags: res.locals.flags
+	};
+
+	getUserFollowed(userId)
```
phamann: Why do you still go and fetch the user follows when we have that data next-api response?

server/controllers/following/layout.js
```diff
@@ -1,56 +1,60 @@
...
+		.then(followedStuff => Object.assign(model, followedStuff))
```
phamann: Are you intentionally mutating `model` here? May be safe to use `Object.assign({}, model, followedStuff)`

server/lib/content-decorators/following-inbox.js
```diff
@@ -0,0 +1,19 @@
+const {fetchContent} = require('../next-api-client');
+const saveButtonTemplate = require('@financial-times/n-ui/myft/templates/save-for-later.html');
+
+function addSaveButtons (content, flags) {
+	return content.map(c => Object.assign({}, c, {
+		actions: (content.actions || []).concat([saveButtonTemplate({contentId: c.id, flags})])
+	}));
```
phamann: You can lose the `Object.assign` and just wrap object literal in parens `()`

server/lib/next-api-client.js
```diff
@@ -0,0 +1,34 @@
...
+				});
+			}
+			return res.json();
+		})
+}
+
+
+module.exports = {
+	fetchContent: userId => makeNextApiCall(userContent, {uuid: userId})
+			.then(({ data = {} } = {}) => data.user.myFtFeed)
```
phamann: Do you really want the default params here? That could mean that we send empty data to the user?

PR #1011, opened 2016-11-10
Following cards next api

laurieboyes:
A continuation of this PR https://github.com/Financial-Times/next-myft-page/pull/1009, this time extending to the following page topic view.

Behind the `myFtPageNextApi` flag

Note that this PR also splits out the lists and followed cards' templates and view models because I think any benefit we were getting from sharing that code was outweighed by all the messy conditionals it required.


server/controllers/following/layout.js
```diff
@@ -8,7 +8,7 @@ module.exports = (req, res, next) => {
 
 	// if myFtPageNextApi flag is off or view isn't time (temp), revert to legacy version
 	// todo after next-api-refactor: remove
-	if(!res.locals.flags.myFtPageNextApi || req.params.layout !== 'time') {
+	if(!res.locals.flags.myFtPageNextApi || !(req.params.layout === 'time' || req.params.layout === 'topic')) {
```
keirog: Intrigued by this but will chat offline

laurieboyes: It's just because I'm scared of messing up .rss/.json (if we even need json anymore which I don't think we do)

keirog: Ah, okies ta

PR #1015, opened 2016-11-11
Handlebars

debugwand:
Handlebars for following page, and some comments for us to track usage of n-section and n-card


server/models/topic-cards.js
```diff
@@ -5,7 +5,7 @@ const tap = require('../lib/utils/tap');
 const podcastIdV1 = 'NjI2MWZlMTEtMTE2NS00ZmI0LWFkMzMtNDhiYjA3YjcxYzIy-U2VjdGlvbnM=';
 const propertyEquals = (property, value) => item => item[property] === value;
 
-import { ConceptCard } from '@financial-times/n-card';
+import { ConceptCard } from '@financial-times/n-card';//TODO: what is this doing here?
```
keirog: Looks like it was needed in order to pass it through to the `reactRenderToString` helper in the `followed-topic.html` template. Not sure if that answers your question

PR #1043, opened 2017-01-14
a11y: low myFT card contrast when removed

lc512k:
Since users can still interact with cards that have been removed, they're expected to have contrast above 4.5:1 (flagged by DAC for a11y accreditation). Reducing the opacity to .5 breaks that rule.

~~This is one alternative. Happy with any other that preserves contrast, pending a potential redesign.~~

ETA: Removed fade-out completely after discussion with Tricia/Sue/Keith, as interim solution pending a redesign of this feature

@debugwand @keirog @jkerr321


keirog:
:wilted_flower: Farewell, old fader

keirog:
cc @Financial-Times/next-myft

laurieboyes:
Bittersweet code deletion 👍

PR #1048, opened 2017-01-24
Replace old v1 params

debugwand:


server/models/article-v1.js
```diff
@@ -16,7 +16,7 @@ module.exports = function(article) {
 	if (article.item.lifecycle.lastPublishDateTime) {
 		output.lastUpdated = article.item.lifecycle.lastPublishDateTime;
 	}
-	output.subheading = article.item.editorial.standFirst
+	output.standFirst = article.item.editorial.standFirst
```
laurieboyes: Sorry, bit late, but is this camelCase intentional?

laurieboyes: (also sorry this is such a mess - this file is one of the ones that will die with personalised feed)

laurieboyes: Ah right actually we're probably not using that property anywhere 👍

debugwand: no. I think this file is unused anyway but will update

PR #1075, opened 2017-03-01
add alerts v2 controller and vary on flag

leggsimon:


server/lib/vary-controller-on-flag.js
```diff
@@ -0,0 +1,6 @@
+const varyControllerOnFlag = (flag, controllerWhenTrue, controllerWhenFalse) => (req, res, next) => {
+	const flags = res.locals.flags;
+	return flags[flag] ? controllerWhenTrue(req, res, next) : controllerWhenFalse(req, res, next);
```
GlynnPhillips: Just to safeguard for the unlikely case where `res.locals.flags` didn't exist it might be worth adding a check for flags first in the condition: `flags && flags[flag] ? controllerWhenTrue(req, res, next) : controllerWhenFalse(req, res, next)`

views/alerts-v2.html
```diff
@@ -0,0 +1,3 @@
+<div class="mypage" id="site-content" role="main">
+	Hello, {{{userId}}}!
+</div>
```
GlynnPhillips: This is a take it or leave it comment as its just my personal opinion but I think this would be better off inside a directory like `views/alerts-v2/alerts.html` My reasons are

1) It mirrors the controllers so to me just makes it easier to follow the code around
2) We can move everything we need into the new directory, even if it means some duplication, then it should make it easier to tidy up old unused code when we have fully switched over

test/server/controllers/alerts-v2.spec.js
```diff
@@ -0,0 +1,31 @@
+const app = require('../../../server/init');
```
GlynnPhillips: Again similar to this comment https://github.com/Financial-Times/next-myft-page/pull/1075/files#r103882385

I prefer using directories to split up files than namespacing

PR #1076, opened 2017-03-01
don't call `build-production` within `make test`

leggsimon:
don't call `build-production` within `make test`


laurieboyes:
Haha what???

PR #1081, opened 2017-03-03
alerts page handles email platform not knowing anything about user

laurieboyes:


laurieboyes:
Don't be hasty - fix in a more considered way on the new one

PR #1082, opened 2017-03-06
Preference page sections / headings

GlynnPhillips:


debugwand:
1) name other than section so people don't conflate this with n-section
2) folder location as a direct child of partials as we will use this approach on all our pages
3) get 'template' out of server-side like this: `{{contents templatename data}}`
4) put title in `contents` and use template strings so html gets properly handled. linter might complain but we really should template strings for anything with unescaped html in it
```
{
  uuid: '1234',
  title: 'page title',
  contents: {
    instant: {
      title: `<span aria-label="my F T">myFT</span> Instant Alerts`,
      etc
    },
   digest: {
      title: `<span aria-label="my F T">myFT</span> Digest`,
      etc
    }
  }
}
```

PR #1088, opened 2017-03-08
Alerts - all models return promises. Also boilerplate for newsletters bit.

laurieboyes:
Whoops! Got carried away a bit. Does the thing we've talked about where models return promises by default, and fixes all the tests.

Also a bit of the newsletters stuff muddled in (sorry)


test/server/controllers/v2/alerts/read.spec.js
```diff
@@ -60,12 +63,10 @@ describe('Alerts controller', () => {
 			}
 		};
 		stubs.getUserInformationStub.returns(Promise.resolve(userInfo));
-		stubs.getNavigationModelStub.returns({});
-		stubs.digestModelStub.returns({});
-		stubs.instantModelStub.returns({});
 
 		return controller({}, res).then(() => {
-			expect(res.render).to.have.been.calledWith('v2/alerts', viewModel);
+			expect(res.render.getCall(0).args[0]).to.deep.equal('v2/alerts');
```
laurieboyes: This stuff where I've swapped `calledWith` for doing a deep.equal on the specific call args looks weird but I find it gives much more helpful test fail messages. E.g. the difference between:

screen shot 2017-03-08 at 17 48 59

and 

screen shot 2017-03-08 at 17 49 18

Compromises on both sides though so happy to discuss if people prefer it some other way.

PR #1089, opened 2017-03-08
WIP: Myft digest panel

leggsimon:
Ok, so this is the first pass at the digest box and instant alerts box, it passes through the topics and preferences to the template. I have wired up unsubscribe buttons for enhanced browsers but not for core yet but will do that in another PR to keep them down in size

### Todo
- [x] Wire up digest unsubscribe buttons
- [x] Wire up digest subscribe buttons
- [ ] Wire up preview buttons
- [ ] Handle rejected promises in the digest model
- [ ] Handle rejected promises in the instant model
- [ ] Write up instant alert model
- [ ] Style inline with new designs
- [ ] Move different scenarios in `digest.html` into separate partials (_under consideration_ 🤔 )
- [x] 🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻🍻  

-----------------


client/components/v2/alerts/main.scss
```diff
@@ -47,3 +47,12 @@ $o-teaser-collection-is-silent: false;
 	@include oIconsGetIcon('edit', oColorsGetPaletteColor('teal-1'), 20);
 	margin-bottom: -5px;
 }
+
+p.alerts-body-copy {
```
laurieboyes: Ah, just spotted this. I don't think we should scope these styles to the `p` elem. Classes will do fine.

leggsimon: Ahh, yes I didn't necessarily mean to keep it like that when I wrote it but thanks for the reminder

laurieboyes: bump

views/partials/v2/alerts/digest.html
```diff
@@ -1 +1,27 @@
...
+	<p class="alerts-body-copy">You get a <span class="alerts-digest__frequency--inline">{{{preferences.type}}}</span> email with the latest stories on all these topics:</p>
+	{{> v2/alerts/topics-flat-list topics}}
+	<p class="alerts-digest__frequency">{{{preferences.type}}}</p>
+	<div class="alerts-email-actions">
+		<button class="o-buttons o-buttons--big alerts__unsubscribe-button" type="button" name="button">Unsubscribe</button>
+		<a class="alerts__email-preview" href="#" aria-label="Preview your digest email">Preview</a>
+	</div>
+{{/if}}
+
+{{#if followsNoTopics}}
```
pixelandpage: Yup, really nice. You'll need to wrap the button in a <form> for core when you come to that bit.

PR #1090, opened 2017-03-09
Alerts hub page gracefully handles errors from email webservices API

laurieboyes:
For rare* occasion when email platform is missing a user, or in the case of any other weird error from the email platform, display a nice error bit under the newsletters heading and don't bring down the whole alerts page.

(*once every three months or so, we're told)

screen shot 2017-03-09 at 12 09 51


views/partials/v2/section.html
```diff
@@ -1,4 +1,4 @@
-<section class="myft-preferences__section" data-trackable="">
+<section class="myft-preferences__section" data-trackable="{{template}}">
```
laurieboyes: Whoops! Bit cheeky using the template string as the `data-trackable` for the section but for as long as they're only one word long, we'll get away with it. Cba to pass a whole new argument for it. Too yolo? :skier:

leggsimon: Hmmm, IMO it's too yolo. In other places we do pass through a `trackable-page` property so I think we should do the same here. For someone writing a new one they might not realise that it would be used for this. You. Stupid. Dick!

test/server/models/v2/alerts/newsletters.spec.js
```diff
@@ -6,27 +6,65 @@ const { expect } = require('chai').use(sinonChai);
...
+
+		return newslettersModel('some-user-id')
+			.then(response => {
+				expect(response.unknownError).to.be.true;
+
+				expect(stubs.nLoggerErrorStub).to.have.been.calledWithMatch('Error getting user\'s newsletters',
+					{
+						event: 'UNKNOWN_ERROR_GETTING_USER_NEWSLETTERS',
+						userId: 'some-user-id'
+						// err: 'Error: Whoa this doesn\'t usually happen', // note - it should add this as well but the error object doesn't sinon.match nicely
```
leggsimon: Remove the comment

server/models/v2/alerts/newsletters.js
```diff
@@ -1,10 +1,21 @@
...
+			logger.error('User not found in email platform', {event:'USER_NOT_FOUND_IN_EMAIL_PLATFORM', userId})
+			return {
+				userMissingInEmailPlatform: true
+			}
+		} else {
+			logger.error('Error getting user\'s newsletters', {event:'UNKNOWN_ERROR_GETTING_USER_NEWSLETTERS', userId, err})
+			return {
+				unknownError: true
+			}
+		}
```
leggsimon: If this throws and is caught here we resolve with either
```js
{ userMissingInEmailPlatform: true }
```
or
```js
{ unknownError: true }
```
right?
If so, how does it render the title heading? Because that is expecting a `title: 'Newsletters'` property no?

test/server/models/v2/alerts/newsletters.spec.js
```diff
@@ -6,27 +6,65 @@ const { expect } = require('chai').use(sinonChai);
...
 
 	it('should return the correct title', function () {
-		return newslettersModel()
+		return newslettersModel('some-user-id')
 			.then(response => {
 				expect(response.title).to.equal('Newsletters')
 			});
 	});
+
+	it('should add a flag to the model if the user is not found by the email webservices API', function () {
```
leggsimon: Super minor and superficial but can we use arrow functions?

test/server/models/v2/alerts/newsletters.spec.js
```diff
@@ -6,27 +6,65 @@ const { expect } = require('chai').use(sinonChai);
...
+				expect(response.userMissingInEmailPlatform).to.be.true;
+
+				expect(stubs.nLoggerErrorStub).to.have.been.calledWith('User not found in email platform', {
+					event: 'USER_NOT_FOUND_IN_EMAIL_PLATFORM',
+					userId: 'some-user-id'
+				})
+			});
+	});
+
+	it('should add a flag to the model if there is an unexpected error from the email webservices API', function () {
```
leggsimon: `=>`

views/partials/v2/alerts/newsletters.html
```diff
@@ -1 +1,9 @@
-<p>Read all the pamphlets</p>
+{{#if userMissingInEmailPlatform}}
+	<p class="alerts-body-copy" data-trackable="user-missing-error">Sorry, there was a problem getting newsletters for your account. Please get in touch with <a href="http://help.ft.com/" data-trackable="customer-support">Customer Support</a>.</p>
+{{else if unknownError}}
```
leggsimon: Woah! Does handlebars have `else if`??

test/server/models/v2/alerts/newsletters.spec.js
laurieboyes: Sure why not 👍

views/partials/v2/alerts/newsletters.html
laurieboyes: Yeah since 3.0 apparently 🎉

server/models/v2/alerts/newsletters.js
laurieboyes: image

Thanks

server/models/v2/alerts/newsletters.js
```diff
@@ -1,10 +1,27 @@
...
+
 module.exports = userId => getUserNewsletters(userId)
-	.then(newsletters => ({
-		title: 'Newsletters',
+	.then(newsletters => Object.assign({
 		newsletters
-	}))
-	.catch(() => {
-		//todo code and test case where I just 404 or whatever
+	}, defaultProps))
```
leggsimon: This will return an object like this:
```js
{
  title: 'Newsletters',
  newsletters: //something, I don't know what it returns
}
```

Is that right? Do we just want the two top level properties on this object? Maybe we do, I don't know what the result of `getUserNewsletters()` looks like... which brings me onto another point which is, and I'm sorry, can we have a test where the `getUserNewsletters` stub returns a true-to-life response to ensure that it the model returns the correct data given a good response.

test/server/models/v2/alerts/newsletters.spec.js
```diff
@@ -6,27 +6,70 @@ const { expect } = require('chai').use(sinonChai);
...
+
+	it('should add a flag to the model if the user is not found by the email webservices API', () => {
+
+		stubs.getUserNewslettersStub.returns(Promise.reject(new Error('404')));
+
+		return newslettersModel('some-user-id')
+			.then(response => {
+				expect(response.userMissingInEmailPlatform).to.be.true;
+
+				// it should still have the title
```
leggsimon: remove this comment. Rather than asserting two different parts of it maybe it would be better to just assert the response is deeply equal to an expected object like so
```js
const expected = {
  title: 'Newsletters',
  userMissingInEmailPlatform: true
};

return newslettersModel('some-user-id')
  .then(response => {
    expect(response).to.eql(expected);
})
```

test/server/models/v2/alerts/newsletters.spec.js
```diff
@@ -6,27 +6,70 @@ const { expect } = require('chai').use(sinonChai);
...
+	it('should add a flag to the model if there is an unexpected error from the email webservices API', () => {
+
+		stubs.getUserNewslettersStub.returns(Promise.reject(new Error('Whoa this doesn\'t usually happen')));
+
+		return newslettersModel('some-user-id')
+			.then(response => {
+				expect(response.unknownError).to.be.true;
+
+				// it should still have the title
+				expect(response.title).to.equal('Newsletters');
```
leggsimon: Ditto

PR #1092, opened 2017-03-09
Include the main image in the response

GlynnPhillips:
This can be used by the push notifications to display the main image
instead of the myFT icon


leggsimon:
🤷‍♂️ sure

PR #1097, opened 2017-03-10
sort out the section dividing lines

leggsimon:
screenshot 2017-03-12 10 25 17

In order to have the "Unsubscribe" button and "Preview" link on the same line as the section heading I've had to remove proper usage of `o-teaser-collection` as the heading text was forced to be `display: block;` and changing it to `inline` forced the bar to only be as wide as the text. In order to get around this I've had to put the styling for the bar on the sections' `::before` and `::after` elements.

cc/ @onishiweb I had Glynn's help on this and he suggested letting you know of our use case of having additional elements inline with the heading text. Unfortunately I've had to copy the scss [from `o-teaser-collection`](https://github.com/Financial-Times/o-teaser-collection/blob/master/src/scss/_mixins.scss) rather than using the mixins as they also add the text styling which I didn't want. I wonder whether those mixins could be split out so that we could just add a mixin on an element that gives it the `::before` and `::after` styles but doesn't add any font styling?


client/components/v2/alerts/heading.scss
```diff
@@ -0,0 +1,44 @@
...
+		border-top: 8px solid rgba(oColorsGetPaletteColor('black'), 0.05);
+		content: '';
+		left: 0;
+		top: 0;
+		position: absolute;
+		width: 100%;
+	}
+
+
+	&:after {
```
GlynnPhillips: Not sure how sass will convert this but we should probably use the CSS3 syntax with double colon `::after`

views/v2/alerts.html
```diff
@@ -2,7 +2,7 @@
 	<div class="o-grid-container" data-trackable="alerts">
 		<header class="alerts-header o-grid-row">
 			<div data-o-grid-colspan="12">
-				<h2 class="alerts__title">Emails &amp; Alerts</h2>
+				<h1 class="alerts__title" aria-label="">Emails &amp; Alerts<span class="n-util-visually-hidden"> Preferences</span></h1>
```
debugwand: aria label should not be empty

views/partials/v2/alerts/instant.html
```diff
@@ -1,10 +1,10 @@
 {{#if hasInstantAlerts}}
-	<p class="alerts-body-copy">You get an Instant Alert email with the latest stories on all these topics, visit the myFT following page to edit these topics:</p>
-	{{> v2/alerts/topics-flat-list topics}}
 	<div class="alerts-email-actions">
 		<button class="o-buttons o-buttons--big alerts__unsubscribe-button" type="button" name="button">Unsubscribe</button>
 		<a class="alerts__email-preview" href="#" aria-label="Preview your digest email">Preview</a>
 	</div>
+	<p class="alerts-body-copy">You get an Instant Alert email with the latest stories on all these topics, visit the myFT following page to edit these topics:</p>
```
debugwand: is class name on every paragraph because of the size difference? suggest we override paragraph styles for the page rather than a class on every element

leggsimon: Yes, I added that class for the weird body copy styles with the 60% grey colour and some bespoke styles. I don't mind overriding `p` elements but think that's generally bad practise although I have no opinion on it either way

client/components/v2/alerts/heading.scss
leggsimon: Are you updating this or shall I?

debugwand: no rush, it can wait until tomorrow when you are back in

client/components/v2/alerts/heading.scss
```diff
@@ -0,0 +1,43 @@
...
+	margin: 0 15px 15px 0;
+	display: inline-block;
+}
+
+.myft__section-heading-actions {
+	display: inline-block;
+	margin-bottom: 10px;
+}
+
+.myft__section-heading-actions--button {
```
GlynnPhillips: The naming of this and the way it is used looks more like an  _element_ than a _modifier_. 

Do we need the `myft` prefix? did you find this was needed to prevent style bleeding? I am just wondering because at the moment it is taking on the _block_ role which I think is partly why we have ended up with `--button` taking on the _modifier_  role as everything is being shifted on one position.

Personally, I think this would be nicer becoming something like `section-heading-actions__button`

If we really need to prefix our styles have we considered doing it via a body class? For instance `.myft .section-heading-actions__button`

PR #1103, opened 2017-03-15
style the column content

leggsimon:
This adds the styling for the column content, you can use this by wrapping the content in a containing `div.myft-section__column-content`. Have fun !! 🎉 

### Here, have some images
----------------------

#### < M
screenshot 2017-03-15 11 54 30

#### M < XL
screenshot 2017-03-15 11 54 45

#### > XL
screenshot 2017-03-15 11 54 58


GlynnPhillips:
😍 screenshots

client/components/v2/alerts/section.scss
```diff
@@ -51,3 +51,14 @@
...
 }
+
+.myft-section__column-content {
+	@include oGridColspan((default: full-width, M: 8, XL: 6));
+
+	@include oGridRespondTo(M) {
+		@include oGridCenter;
+	};
+
+	float: none;
```
GlynnPhillips: Do we need to float? I would have thought o-grid handled this but maybe I am wrong

debugwand: are we resetting styles? can we avoid doing that?

leggsimon: Maybe there's a better way of doing it but with `float` this happens on full-width
screenshot 2017-03-15 13 56 49
Removing the `float` property seems to fix that.

PR #1106, opened 2017-03-15
Unsubscribe from all

pixelandpage:
screen shot 2017-03-15 at 16 08 06
screen shot 2017-03-15 at 16 08 31

Still to do:
- [ ] Flesh out the unsubscribe logic, currently only checks for an email preference
- [ ] Build the enhanced journey
- [ ] Two-state button
- [ ] Make responsive / mobile view
- [ ] Confirm colours and typography (o-colors beta)


client/components/v2/alerts/unsubscribe-from-all.scss
```diff
@@ -0,0 +1,25 @@
+.alerts__unsubscribeFromAll-container {
```
leggsimon: These should be updated to use our new BEM convention. I'll try and think of some appropriate names. 🤔 
 https://docs.google.com/document/d/1uGDymFhJiUWK2QxXeWzdlpv0t5VyvliXZQdK2WAhrA0/

client/components/v2/alerts/unsubscribe-from-all.scss
```diff
@@ -0,0 +1,25 @@
...
+	display: flex;
+	align-items: center;
+	justify-content: center;
+}
+
+.alerts__unsubscribeFromAll-button {
+	@include oTypographySansBold(m);
+	color: oColorsGetPaletteColor('white');
+	padding: 8px;
+	font-size: 20px;
```
leggsimon: `oTypographySansBold(m)` already has a 20px `font-size` so you shouldn't need to have this. Having said that. Becky was reducing some 20px font-sizes down to 18px. So either remove this or keep it and change to 18px.

client/components/v2/alerts/unsubscribe-from-all.scss
debugwand: see css doc - will also come over

client/components/v2/alerts/unsubscribe-from-all.scss
```diff
@@ -0,0 +1,25 @@
...
+	background-color: oColorsGetPaletteColor('black');
+	border: 1px solid oColorsGetPaletteColor('black');
```
leggsimon: this might be a stupid comment but if the whole button is black does it need to also have a black border?

client/components/v2/alerts/unsubscribe-from-all.scss
```diff
@@ -0,0 +1,25 @@
+.alerts__unsubscribeFromAll-container {
+	background-color: oColorsGetPaletteColor('red-tint5');
+}
+
+.alerts__unsubscribeFromAll-message {
+	@include oTypographySans(m);
+	color: oColorsGetPaletteColor('black');
+	margin: 25px;
```
leggsimon: Not entirely sure, maybe someone else does, but does this text want margin or should the container have padding?

server/models/v2/alerts/unsubscribeFromAll.js
```diff
@@ -0,0 +1,7 @@
+const fetchUserPrefs = require('../../../lib/fetch-preferences');
+
+module.exports = (uuid) => fetchUserPrefs(uuid)
+	.then((userPrefs) => {
+		const canUnsubscribeFromAll = Boolean(userPrefs['email-digest']);
+		return { canUnsubscribeFromAll }
```
leggsimon: I wonder if this should just be simplified to 
```js
return {
    canUnsubscribeFromAll = Boolean(userPrefs['email-digest'])
}
```

test/server/models/v2/alerts/unsubscribe-from-all.spec.js
```diff
@@ -0,0 +1,43 @@
...
+
+describe('Alerts unsubscribe from all model', () => {
+	afterEach(() => {
+		sandbox.restore();
+	});
+
+	const unsubscribeFalse = {}
+
+	const unsubscribeTrue =
+	{ 'email-digest':
```
leggsimon: curly brace should start on the above line

test/server/models/v2/alerts/unsubscribe-from-all.spec.js
```diff
@@ -0,0 +1,43 @@
...
+		{ name: 'Digest email', uuid: 'email-digest' },
+	};
+
+	it('canUnsubscribeFromAll should be true if userPrefs contains an email digest', () => {
+		stubs.fetchUserPrefsStub.withArgs().returns(Promise.resolve(unsubscribeTrue))
```
leggsimon: Unless you're using `.withArgs()` I'd just get rid of it and have 
```js
stubs.fetchUserPrefsStub.returns(Promise.resolve(unsubscribeTrue))
```

------------------------

Also, although this is personal preference I would either define the response you want it to return inside the `it` block or just have that object directly in the `Promise.resolve` rather than use a variable at all, since it isn't a big object.

test/server/models/v2/alerts/unsubscribe-from-all.spec.js
```diff
@@ -0,0 +1,43 @@
...
+
+		return unsubscribeFromAllModel('uuid').then((response) => {
+			expect(response.canUnsubscribeFromAll).to.be.true
+		})
+	});
+
+	it('canUnsubscribeFromAll should be false if userPrefs doesn\'t contain an email digest', () => {
+		stubs.fetchUserPrefsStub.withArgs().returns(Promise.resolve(unsubscribeFalse))
```
leggsimon: ditto

server/models/v2/alerts/unsubscribeFromAll.js
pixelandpage: Seems not to be supported

debugwand: well the equals would need to be a colon :)

pixelandpage: Magic!

leggsimon: 🤦‍♂️

client/components/v2/alerts/unsubscribe-from-all.scss
```diff
@@ -0,0 +1,24 @@
+.myft-callout__container {
```
GlynnPhillips: Does this need to be scoped as an _element_ looking at the markup I think it could just be `myft-callout` as the _block_

test/server/models/v2/alerts/unsubscribe-from-all.spec.js
```diff
@@ -0,0 +1,43 @@
...
+
+describe('Alerts unsubscribe from all model', () => {
+	afterEach(() => {
+		sandbox.restore();
+	});
+
+
+const unsubscribeTrue = {
+		'email-digest': {
+			name: 'Digest email', uuid: 'email-digest'
```
GlynnPhillips: Could we put these on new lines please?

```
name: 'Digest email', 
uuid: 'email-digest'
```

views/partials/v2/alerts/unsubscribeFromAll.html
```diff
@@ -0,0 +1,14 @@
+<section class="myft-preferences__section" data-trackable="{{trackable}}">
+	<div class="o-grid-container myft-callout__container">
```
GlynnPhillips: As per the comment in the css I am not sure this needs the `__container` part

views/partials/v2/alerts/unsubscribeFromAll.html
```diff
@@ -0,0 +1,14 @@
+<section class="myft-preferences__section" data-trackable="{{trackable}}">
+	<div class="o-grid-container myft-callout">
+		<div class="o-grid-row">
+			<div data-o-grid-colspan="10">
+					<p class="myft-callout__message">By unsubscribing from all emails listed on this page, you will no longer receive any myFT Digests, Instant Alerts, Newsletters or Updates and Offers from the FT.</p>
+			</div>
+			<div data-o-grid-colspan="2" class="myft-callout__button-wrapper">
+				<form class="unsubscribeFromAll__form" method="post" action="/__myft/api/alerts/{{@root.userId}}/unsubscribe">
```
laurieboyes: Any reason for the case changing here? Should `unsubscribeFromAll__form` be `unsubscribe-from-all__form `

views/partials/v2/alerts/unsubscribeFromAll.html
```diff
@@ -0,0 +1,14 @@
...
+					<button class="o-buttons o-buttons--inverse o-buttons--big o-buttons--big myft-callout__button" type="submit" name="button" data-trackable="alerts__unsubscribeFromAll-button" {{#unless canUnsubscribeFromAll}}disabled{{/unless}}>Unsubscribe from all</button>
```
laurieboyes: Trackable value of `alerts__unsubscribeFromAll-button ` <- doesn't need to follow any bem conventions or anything. Will get the fact that it's alerts from the trackable tree thing it gets from it's ancestor elements (see the newsletter unsubscribe path below for an example).

Maybe simplify this attribute to just `unsubscribe?`

screen shot 2017-03-17 at 15 48 23

PR #1110, opened 2017-03-20
use sandbox to clean up stubs after each test

leggsimon:
@debugwand I think this was the one you mentioned earlier


debugwand:
see https://github.com/Financial-Times/next-myft-page/blob/master/test/server/controllers/v2/alerts/unsubscribe-via-get.spec.js and http://sinonjs.org/releases/v2.0.0/sandbox/

stubs need to be created in beforeEach

leggsimon:
Does this matter since the stubs were being set up in each test anyway?

I have just refactored the repeated stubs to be shared between tests, setting up and restoring each of them before and after respectively. But I think this is more just a tidying up rather than whether it was actually necessary?

debugwand:
the cod tidy-up is fine, the comment was to point to how we work with restore not resetting stubs, as per our discussion yesterday

leggsimon:
Yep I realised after I wrote that comment that it was much more pertinent to the changes I've just made in `digest.spec.js` as that had stubs that were being set up once and being shared amongst all the individual tests.

PR #1113, opened 2017-03-21
Core digest settings form

leggsimon:
adds a _relatively_ unstyled core form

screenshot 2017-03-21 12 16 25

Todo (possibly in a different PR)

- [ ] hide submit buttons when enhanced
- [ ] style the radio buttons if possible
- [ ] decide whether radio buttons and a submit was the right thing to do in the first place or whether buttons would be better
- [ ] contemplate life and own existence
- [x] add mixin for the `$x-percent-black` scss variables


laurieboyes:
> contemplate life and own existence
I'd steer clear of this one mate

views/partials/v2/alerts/digest.html
```diff
@@ -8,6 +8,23 @@
...
+						<input type="radio" name="_rel.type" id="digest-frequency-weekly" value="weekly"><label class="myft-body-copy" for="digest-frequency-weekly">Weekly</label>
+						<button class="o-buttons" type="submit">Submit</button>
+					</form>
+					<form class="myft-digest-preferences__setting-option" action="/__myft/api/core/preferred/preference/email-digest?method=put" method="post">
+						<legend class="myft-section-heading--small">Choose Email Timezone</legend>
+						<input type="radio" name="_rel.timezone" id="digest-timezone-europe" value="Europe/London"><label class="myft-body-copy" for="digest-timezone-europe">Europe &amp; Africa</label>
+						<input type="radio" name="_rel.timezone" id="digest-timezone-america" value="America/New_York"><label class="myft-body-copy" for="digest-timezone-america">America</label>
+						<input type="radio" name="_rel.timezone" id="digest-timezone-asia" value="Asia/Hong_Kong"><label class="myft-body-copy" for="digest-timezone-asia">Asia</label>
+						<button class="o-buttons" type="submit">Submit</button>
```
laurieboyes: Could we change this to 'Update' and then for screen-reader users we could have it say 'Update digest timezone' using hidden text?

laurieboyes: (or something)

leggsimon: Yep, guess so

views/partials/v2/alerts/digest.html
```diff
@@ -8,6 +8,23 @@
 			<p class="myft-body-copy">You get a <span class="alerts-digest__frequency--inline">{{{preferences.type}}}</span> email with the latest stories on all these topics:</p>
 			{{> v2/alerts/topics-flat-list topics}}
 			<p class="alerts-digest__frequency">{{{preferences.type}}}</p>
+			<div class="myft-digest-preferences">
+				<div class="myft-digest-preferences__setting">
+					<form class="myft-digest-preferences__setting-option" action="/__myft/api/core/preferred/preference/email-digest?method=put" method="post">
+						<legend class="myft-section-heading--small">Choose Email Frequency</legend>
```
pixelandpage: Should we use <label> rather than <legend>? The latter expects <fieldset> as its parent https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend. Also I suspect that <label> will be better supported for a11y.

views/partials/v2/alerts/digest.html
```diff
@@ -8,6 +8,23 @@
...
+						<input type="radio" name="_rel.type" id="digest-frequency-daily" value="daily"><label class="myft-body-copy" for="digest-frequency-daily">Daily</label>
+						<input type="radio" name="_rel.type" id="digest-frequency-weekly" value="weekly"><label class="myft-body-copy" for="digest-frequency-weekly">Weekly</label>
+						<button class="o-buttons" type="submit">Submit</button>
+					</form>
+					<form class="myft-digest-preferences__setting-option" action="/__myft/api/core/preferred/preference/email-digest?method=put" method="post">
+						<legend class="myft-section-heading--small">Choose Email Timezone</legend>
```
pixelandpage: As above

PR #1115, opened 2017-03-21
Don't included automated newsletters in new alerts page

laurieboyes:


server/models/v2/alerts/newsletters.js
```diff
@@ -10,6 +10,7 @@ module.exports = userId => getUserNewsletters(userId)
 		return Object.assign({
 			newsletters: newsletters
 				.filter(newsletter => !newsletter.inactive)
```
debugwand: marginally better performance if we do both tests in 1 filter statement

laurieboyes: Yeah I guess but I feel like two lines is marginally easier to read and I like the idea of it being a single whole-line delete if the functionality changes. We must pick our favourite margin. Happy to be overruled here.

leggsimon: My personal preference would be to side with Laurie purely because I find it easier to read. I wonder if the performance gain is so negligible considering it's only like 30 elements?

laurieboyes: Gonna merge for now but happy for debate to continue

PR #1116, opened 2017-03-22
Styling for newsletter cards

laurieboyes:
So _stylish_

Mobile:
screen shot 2017-03-22 at 09 35 42

Desktop:
screen shot 2017-03-22 at 09 35 09

Todo in subsequent PR(s):

- [ ] Apparently this should be using a different version of the financier font


client/components/v2/alerts/main.scss
```diff
@@ -33,7 +34,7 @@ $o-teaser-collection-is-silent: false;
 .myft-body-copy {
 	@include oTypographySans(m);
 	font-size: 18px;
-	color: percentageColor('black', 'pink', 60%);
+	color: percentageColor('black', 'pink', 80%);
```
leggsimon: Has the design changed?

client/components/v2/alerts/newsletter.scss
```diff
@@ -0,0 +1,72 @@
...
+	height: 100px;
+
+	// cancel-out o-grid margin for mobile card
+	@include oGridRespondTo($until: S) {
+		margin-left: -$o-grid-gutter-small;
+		margin-right: -$o-grid-gutter-small;
+	}
+}
+
+.myft-newsletter__topper__heading {
```
leggsimon: `.myft-newsletter__topper-heading`

client/components/v2/alerts/newsletter.scss
```diff
@@ -0,0 +1,72 @@
...
+	// 😢
+	font-size: 30px;
+	line-height: 30px;
+	@include oGridRespondTo($until: S) {
+		font-size: 48px;
+		line-height: 48px
+	}
+}
+
+.myft-newsletter__topper__meta {
```
leggsimon: `myft-newsletter__topper-meta`

client/components/v2/alerts/newsletter.scss
```diff
@@ -0,0 +1,72 @@
...
+	padding-bottom: 20px;
+	margin-bottom: 28px;
+	display: flex;
+	flex-direction: column;
+
+	&::after  {
+		content: '';
+		display: block;
+		margin-top: 21px;
+		border-bottom: 1px solid $twenty-percent-black;
```
leggsimon: Should use `percentageColor('black', 'pink', 20%)`

client/components/v2/alerts/newsletter.scss
```diff
@@ -0,0 +1,72 @@
...
+
+.myft-newsletter__description {
+	flex-grow: 1;
+	margin: 12px 0 0 0;
+}
+
+.myft-newsletter__meta {
+	@include oTypographySans(xs);
+	text-transform: uppercase;
+	color: $sixty-percent-black;
```
leggsimon: Should use `percentageColor('black', 'pink', 60%)`

client/components/v2/alerts/newsletter.scss
```diff
@@ -0,0 +1,72 @@
...
+	}
+}
+
+.myft-newsletter__topper__meta {
+	@include oTypographySansBold(s);
+	position: absolute;
+	top: 0;
+
+	right: 16px;
+	// cancel-out o-grid margin for mobile card
```
leggsimon: I seem to think there's a nicer way to do this... 🤔

views/partials/v2/alerts/newsletters.html
```diff
@@ -5,22 +5,27 @@
...
-				type="submit">Unsubscribe<span class="n-util-visually-hidden">&nbsp;from {{name}}</span></button>
-		</form>
-	</li>
-	{{/newsletters}}
+<ul class="o-grid-container o-grid-container--bleed">
+	<div class="o-grid-row">
+		{{#newsletters}}
+		<li class="myft-newsletter" data-o-grid-colspan="12 S6 M4 L3">
+			<div class="myft-newsletter__topper">
+				<h3 class="myft-newsletter__topper__heading">{{name}}<span class="n-util-visually-hidden">{{#if isPremium}}, premium{{/if}}{{#if frequency}}, {{frequency}}{{/if}}</span></h3>
```
leggsimon: `myft-newsletter__topper-heading`

views/partials/v2/alerts/newsletters.html
```diff
@@ -5,22 +5,27 @@
...
+				{{#if isPremium}}<span aria-hidden="true" class="myft-newsletter__topper__meta o-labels o-labels--premium">Premium</span>{{/if}}
```
leggsimon: `myft-newsletter__topper-meta`

views/partials/v2/alerts/newsletters.html
```diff
@@ -5,22 +5,27 @@
...
+			<p class="myft-newsletter__description myft-body-copy">{{longDesc}}</p>
+			<div aria-hidden="true" class="myft-newsletter__meta">{{frequency}}</div>
+			<form action="{{unsubscribeAction}}" method="POST">
+				<input type="hidden" name="listName" value="{{name}}">
+				<button
+					class="o-buttons o-buttons--big"
+					aria-label="Unsubscribe from {{name}}"
+					title="Unsubscribe from {{name}}"
+					data-trackable="newsletter-unsubscribe"
+					type="submit">Unsubscribe<span class="n-util-visually-hidden">&nbsp;from {{name}}</span></button>
```
leggsimon: Does  `{{name}}` need to be  `{{{name}}}` so that it actually inserts it?

client/components/v2/alerts/main.scss
laurieboyes: Possibly. It's 80 on the current one ¯\_(ツ)_/¯

leggsimon: Hmmmm, It shouldn't be I don't think. Maybe I messed up. Will take a look

client/components/v2/alerts/newsletter.scss
laurieboyes: Yeah I looked into [o-grid and full-bleed containers](https://github.com/Financial-Times/o-grid#full-bleed-container) but don't think we can use them in this case due to the newsletter card being nested inside other containers, and further by the fact that it only wants to be full-bleed on mobile

client/components/v2/alerts/main.scss
leggsimon: Yeah no it is meant to be 60%

views/partials/v2/alerts/newsletters.html
laurieboyes: Don't think so? I thought the 'triple-stash' is just for HTML escaping. Peeking at the markup in devtools suggests this has worked fine.

Is it that I'm causing confusion with the `&nbsp;`? It occurs to me now that I'm not actually sure I need it, I'm just paranoid (possibly for vague personal experience reasons) some browsers will collapse a normal space character at the beginning of the tag and it'll get read out all smushed together.

client/components/v2/alerts/newsletter.scss
debugwand: is it something where an extra origami style would help?

views/partials/v2/alerts/newsletters.html
```diff
@@ -5,22 +5,27 @@
...
-				type="submit">Unsubscribe<span class="n-util-visually-hidden">&nbsp;from {{name}}</span></button>
-		</form>
-	</li>
-	{{/newsletters}}
+<ul class="o-grid-container o-grid-container--bleed">
+	<div class="o-grid-row">
+		{{#newsletters}}
+		<li class="myft-newsletter" data-o-grid-colspan="12 S6 M4 L3">
+			<div class="myft-newsletter__topper">
+				<h3 class="myft-newsletter__heading">{{name}}<span class="n-util-visually-hidden">{{#if isPremium}}, premium{{/if}}{{#if frequency}}, {{frequency}}{{/if}}</span></h3>
```
debugwand: I think this may be more complex than required, and those bits of text don't need to be part of the heading, and premium is made available to screenreaders instead

views/partials/v2/alerts/newsletters.html
debugwand: the space character might get removed by tools like handlebars rather than browsers

views/partials/v2/alerts/newsletters.html
laurieboyes: I went for this solution in the end based on responses to my question in the accessibility channel, which made a good case for including this stuff in the title: https://financialtimes.slack.com/archives/C2LMEKC6S/p1490027049236993

Do you strongly disagree?

I suppose though I can see that it's a bit weird that the premium bit is duplicated so close together, I just wanted to keep the information consistent.

leggsimon:
I also wonder whether the card should just be extracted out as a new component/partial as well so that your current partial just includes
```
{{ # newsletters }}
  {{> newsletter this}}
{{ / newsletters }}
```

client/components/v2/alerts/newsletter.scss
laurieboyes: Hm, yeah it would definitely be nicer from our perspective. I can think of some things that might make it awkward to implement though, such as the variable levels of nesting that might occur. Also it might not want to use negative margins to bring the container to the edges if it's going to work in the general case, as this ain't no good for absolutely positioning elements within it (as evidenced here). Big 'hmmm' here.

views/partials/v2/alerts/newsletters.html
debugwand: I disagree about it being in the h3 in that structurally this information is meta data not the heading of what this block is about. I am also keen to avoid diverging the experience between different user groups - do x for these users and y for these ones - as it adds complexity to our code and means you have to consider 'this happens for these people and that happens for those people' when building, rather than thinking in the spirit of progressively enhancing something.

I do agree frequency should be next to premium, that could be achieved with flexbox.

client/components/v2/alerts/newsletter.scss
```diff
@@ -0,0 +1,69 @@
...
+	display: flex;
+	height: 100px;
+	position: relative;
+	box-sizing: border-box;
+
+	// cancel-out o-grid margin for mobile card
+	@include oGridRespondTo($until: S) {
+		left: -$o-grid-gutter-small;
+		width: calc(100% + #{$o-grid-gutter-small * 2});
+	}
```
laurieboyes: @debugwand Position relative was a much neater solution, and I no longer need to tweak the position of the premium label 🎉 Thank you for prompting me to properly learn how position relative works.

Still not sure whether or not this would be useful as a general use origami class though, mostly due to complications such as the potential for different levels of nesting :thinking:

laurieboyes:
@leggsimon Re the newsletter partial thing. Yeah that's v sensible, but it'll get picked up as part of the newsletter page work for sure so I'll leave it for now.

laurieboyes:
Actually scratch that, I'll do it now then any changes to the component for the newsletter page will be clearer in future commits 👍 nice one

leggsimon:
Also, hmm, annoying that coveralls seems to be counting html and scss files. I see if there's anything we can do

client/components/v2/alerts/newsletter.scss
debugwand: probably not origamifyable, because it is a situation caused by the component inside another one so is not atomic

leggsimon:
Ignore coveralls I think it's checking the additional line in `bower.json`

client/components/v2/alerts/main.scss
laurieboyes: Gonna merge now and clarify with design later 👍

leggsimon: Hokay

PR #1118, opened 2017-03-22
change digest settings onChange of radio buttons

leggsimon:
🤢  I don't feel good about this code, it feels super icky for some reason.

I do still need to think about a11y and screen reading. Wonder if someone might give some guidance there.

Also need to work out a nice way to "check" the current setting on initial load


client/components/v2/alerts/index.js
```diff
@@ -1,24 +1,39 @@
...
 	const defaultRelationshipSettings = {_rel: {timezone: 'Europe/London', type: 'daily'}};
 	myftApiClient.add('user', null, 'preferred', 'preference', 'email-digest', defaultRelationshipSettings);
 }
 
+const updateDigestSetting = (e) => {
+	const name = e.target.getAttribute('name');
+	const value = e.target.getAttribute('value');
+	const key = name.replace(/^_rel\./, '')
+	const updatedRelationshipSettings = { _rel: {[key] : value} }
+	myftApiClient.updateRelationship('user', null, 'preferred', 'preference', 'email-digest', updatedRelationshipSettings)
```
laurieboyes: Could we give some sort of feedback that this request is in progress or if it fails? Don't know if it works in this context, but maybe have it behave like follow buttons and disable it while the request is in flight? Then we can put it back to how it was before if the request fails / times out / whatever.

views/partials/v2/alerts/digest.html
```diff
@@ -14,14 +14,14 @@
 						<label class="myft-section-heading--small" for="digest-frequency-form">Choose Email Frequency</label>
 						<input class="myft-settings__radio-button" type="radio" name="_rel.type" id="digest-frequency-daily" value="daily"><label class="myft-body-copy" for="digest-frequency-daily">Daily</label>
 						<input class="myft-settings__radio-button" type="radio" name="_rel.type" id="digest-frequency-weekly" value="weekly"><label class="myft-body-copy" for="digest-frequency-weekly">Weekly</label>
-						<button class="o-buttons" type="submit">Update<span class="n-util-visually-hidden">Email Frequency</span></button>
+						<button class="o-buttons n-util-hide-enhanced" type="submit">Update<span class="n-util-visually-hidden">Email Frequency</span></button>
```
debugwand: add a space before Email or it will get treated as UpdateEmail

client/components/v2/alerts/index.js
```diff
@@ -1,24 +1,39 @@
 const myftApiClient = require('next-myft-client');
 
-const unsubscribe = () => {
+const unsubscribeFromDigest = () => {
 	myftApiClient.remove('user', null, 'preferred', 'preference', 'email-digest');
```
debugwand: wherever we have an asynchronous call, we need to check for failure. if it fails, we need so show an error message and set the UI back to its 'real' state

e.g. I click on 'weekly', but neo times out. I should see an error message nearby (we should have some messaging ui from Olga & Becky to inform the user of updates anyway) and the checked radio should go back to what I had before. 

For accessibility, the error message should either be inside the label element or attached to the input with aria-describedby - https://www.paciellogroup.com/blog/2015/05/short-note-on-aria-labelledby-and-aria-describedby/ https://www.w3.org/WAI/GL/wiki/Using_ARIA_role_of_alert_for_Error_Feedback_in_Forms

leggsimon: I'll deal with this in another PR where I add some core functionality to unsub/sub buttons

PR #1121, opened 2017-03-23
Enhanced unsubscribe

pixelandpage:
- Unsubscribe From All is now a two-state journey if JS is enabled. 
 - Unsubscribing now triggers a refresh instead of a redirect.
 - Some styling.

Note: I suspect that the unsubscribe code in index.js could be moved to a new home. 
Next priority: Test unsubscribe logic
two-state-unsubscribe


client/components/v2/alerts/index.js
```diff
@@ -9,6 +9,24 @@ const subscribe = () => {
 	myftApiClient.add('user', null, 'preferred', 'preference', 'email-digest', defaultRelationshipSettings);
 }
 
+const unsubscribeFromAllButton = document.querySelector('.myft-callout__button');
+const unsubscribeFromAllConfirm = document.querySelector('.myft-callout__button-confirm');
+const unsubscribeFromAllCancel = document.querySelector('.myft-callout__button-cancel');
```
laurieboyes: Could we add some extra classes expressly for selecting elements with javascript? e.g. `.js-unsub-from-all`, `.js-unsub-from-all-confirm`. That way we have nice separation of concerns between the styling classes and the javascript, and if anyone comes back to this in 3 months time changing the styling and fancies changing some class names, they won't have to worry about breaking the JS side.

client/components/v2/alerts/unsubscribe-from-all.scss
```diff
@@ -1,3 +1,7 @@
+.myft-callout__hide {
```
laurieboyes: Could save a tasty 3 lines and just use `n-util-hide` https://github.com/Financial-Times/n-ui-foundations/blob/9e66637b170f83755c93bbdb8064ce7854ac30d4/util/main.scss

client/components/v2/alerts/index.js
```diff
@@ -21,5 +39,18 @@ export default function () {
...
+		if (unsubscribeFromAllButton) {
+			unsubscribeFromAllButton.addEventListener('click', (click) => {
+				click.preventDefault()
+					confirmStateElements.forEach((el) => {toggleElementVisibility(el)});
+			});
+		}
+
+		if (unsubscribeFromAllCancel) {
+			unsubscribeFromAllCancel.addEventListener('click', () => {
+				confirmStateElements.forEach((el) => {toggleElementVisibility(el)});
```
laurieboyes: Seeing as we do this twice, can we put it in a `toggleShowConfirmStuff` (uh, or something) function

PR #1122, opened 2017-03-23
Fix smoke tests

laurieboyes:
Apparently the `sourcepoint` flag is causing problems


andygnewman:
My understanding is that wherever a `headers` property is set, that you should put in the [standard set of pa11y defaults](https://github.com/Financial-Times/n-makefile/blob/master/config/.pa11yci.js#L23) as well as the custom settings.

Also to cover URLs where you don't have specific headers set up, updating to the latest n.Makefile will bring in the sourcepoint off as part of the standard pa11y set.

PR #1123, opened 2017-03-23
Try and get a consistent approach to spacing under the newsletter section headings

laurieboyes:
Try and move away from using margins on components to set spacing between section heading bits and section content.

E.i. this space:
group 2

Removes `myft-section-heading-actions` container and aligns actions and heading within `myft-section__topper` using flexbox 

Possibly controversial because it requires that the section heading markup is moved from the section partial into the individual sections. This means a small amount of markup duplication but I think it's worth it.

Tried to draw a persuasive diagram of how it's laid out now but it was way too much effort, so please do come and ask me IRL to defend this PR.


debugwand:
we need to ditch usePartial anyway because it cannot work client-side - see https://github.com/Financial-Times/n-handlebars/blob/master/express.js#L38

PR #1130, opened 2017-03-28
reduces font size and truncates long topic labels

leggsimon:
This sets a maximum length on topic labels to deal with horrendously long ones like [`ING UK Real Estate Income Trust Ltd (Old name ...`](https://ft.com/stream/organisationsId/MjIxZDRlOTMtY2IxZC00YjMxLTkzYzktMjkxMDU5MzAwODg1-T04=) (Oh god I even got bored typing it out) which also makes the labels flow a bit nicer with fewer empty gaps at the end of the lines

## Before
screenshot 2017-03-28 22 24 39

## After
screenshot 2017-03-28 22 26 24


client/components/v2/alerts/topics-flat-list.scss
```diff
@@ -10,11 +10,16 @@
 	padding: 5px 10px;
 	margin: 5px 10px 5px 0;
 	white-space: nowrap;
+	max-width: 200px;
+	white-space: nowrap;
+	overflow: hidden;
+	text-overflow: ellipsis;
```
laurieboyes: There's actually an `.n-util-truncate` for this: https://github.com/Financial-Times/n-ui-foundations/blob/9e66637b170f83755c93bbdb8064ce7854ac30d4/util/main.scss#L59-L65
Maybe worth using instead, maybe not /shrug. Probably not.

PR #1132, opened 2017-03-29
Expand unsubscribe logic

pixelandpage:
- Extends the unsubscribe logic to check for instant alerts and nbes
 - Adds tests for the expanded unsubscribe logic
 - Now uses data- attributes as js hooks
 - And up-to-date with Simon's snazzy new controller changes


server/models/v2/alerts/unsubscribe-from-all.js
```diff
@@ -1,7 +1,29 @@
...
+	})
+}
+
+function getEmailPrefs (settings) {
+	return settings['email-digest'];
+}
+
+function getNewsletters (userId) {
+	return getUserNewsletters(userId)
+		.then((hasNewsletters) => {
```
laurieboyes: Seeing as this variable is an array rather than a bool, could we rename it to `newsletters`?

server/models/v2/alerts/unsubscribe-from-all.js
```diff
@@ -1,7 +1,29 @@
...
+			canUnsubscribeFromAll: Boolean(emailPrefsResult || newslettersResult || instantAlertsResult)
+		}
+	})
+}
+
+function getEmailPrefs (settings) {
+	return settings['email-digest'];
+}
+
+function getNewsletters (userId) {
```
laurieboyes: The name `getNewsletters` suggests to me that this function returns a list of newsletters, but if I'm correct, it's actually returning a bool for whether or not the user has newsletters? With this in mind, could we rename it to `getUserHasNewsletters` or something?

server/models/v2/alerts/unsubscribe-from-all.js
```diff
@@ -1,7 +1,29 @@
...
+	return getUserNewsletters(userId)
+		.then((hasNewsletters) => {
+			return Boolean(hasNewsletters[0]);
 		})
-	});
-};
+}
+
+function getInstantAlerts (follows) {
```
laurieboyes: Same as my comment on `getNewsletters`

test/server-unit/models/v2/alerts/unsubscribe-from-all.spec.js
```diff
@@ -1,24 +1,76 @@
-const { expect } = require('chai');
+const proxyquire = require('proxyquire');
+const sinon = require('sinon');
+const sinonChai = require('sinon-chai');
+const { expect } = require('chai').use(sinonChai);;
 
-const unsubscribeFromAllModel = require('../../../../../server/models/v2/alerts/unsubscribe-from-all');
+const sandbox = sinon.sandbox.create();
 
 describe('Alerts unsubscribe from all model', () => {
```
laurieboyes: Good tests!

server/models/v2/alerts/unsubscribe-from-all.js
```diff
@@ -1,7 +1,29 @@
...
+}
+
+function getEmailPrefs (settings) {
+	return settings['email-digest'];
+}
+
+function getNewsletters (userId) {
+	return getUserNewsletters(userId)
+		.then((hasNewsletters) => {
+			return Boolean(hasNewsletters[0]);
```
leggsimon: What is `hasNewsletters[0]`? From the naming of it I think I would expect it already to be a Boolean...?

server/models/v2/alerts/unsubscribe-from-all.js
```diff
@@ -1,7 +1,29 @@
-module.exports = (settings) => {
-	return new Promise((resolve) => {
-		resolve({
-			canUnsubscribeFromAll: Boolean(settings['email-digest'])
+const getUserNewsletters = require('../../../lib/email-web-services-client').getUserNewsletters;
```
leggsimon: I think we are already making this call once in the newsletters model. I wonder if we could get this out so we don't duplicate the API call.

Also I think a sort of convention we have is to use Object destructuring to assign the variable so
to keep it consistent I think it should be
```js
const { getUserNewsletters } = require('../../../lib/email-web-services-client');
```

server/models/v2/alerts/unsubscribe-from-all.js
```diff
@@ -1,7 +1,29 @@
...
+	]).then(([emailPrefsResult, newslettersResult, instantAlertsResult]) => {
+		return {
+			canUnsubscribeFromAll: Boolean(emailPrefsResult || newslettersResult || instantAlertsResult)
+		}
+	})
+}
+
+function getEmailPrefs (settings) {
+	return settings['email-digest'];
+}
```
leggsimon: We have this `getEmailPrefs` call inside a `Promise.all()` array so should this return a Promise? Personally I don't think it _should_ as there's no real need for it to be async but I do understand the neatness of having all the things inside the `Promise.all`

server/models/v2/alerts/unsubscribe-from-all.js
leggsimon: Same as _my_ comment above on `getEmailPrefs`

PR #1133, opened 2017-03-29
add first integration test

leggsimon:


test/server-integration/controllers/v2/alerts/read.js
```diff
@@ -0,0 +1,116 @@
+const sinon = require('sinon');
+const sinonChai = require('sinon-chai');
+const { expect } = require('chai').use(sinonChai);;
+const nock = require('nock');
```
laurieboyes: Ooh, genuinely interested to learn why you chose nock over fetch-mock

laurieboyes: Ah fair enough. Nah looks like it'll do the job fine

